PR #4, opened 2020-08-22
Added Website hosted on GH pages

DrMagPie:
Hi I thought that I would be cool if the FFFF would look like an original one :)
So I have restructured this repo and added website which is exact copy of Wube's blog page ( Hope that they don't mind ....)
take a look at [website](https://drmagpie.github.io/Alt-F4/).

For this to work you will need to enable GH pages in repository setting.
Rest should be pretty clear you can see updated blog post to find out how to add thumbnails, author info and forum links.


ClaudeMetz:
Hi there, love the effort! We already have a task-force on creating the website in the Discord (https://discord.gg/AsXAwyV), and there's sadly been some effort duplication it seems. If you could join us on there, we could work this out together?

PR #23, opened 2020-08-25
Crawlertron

LovelySanta:
crawlertron post, as mentioned before, including gfx


ClaudeMetz:
Alright I did my editing thing on this one. Overall, I think this post is really great, good mix of humor and technical stuff. I also removed some of the introduction because that ground will already be covered by the general introduction to the blog post.

I'm relatively new at this editing thing, so I'm trying to find the balance between making the whole thing flow better and keeping the original intention and character. I changed quite a few things, hopefully you're not too offended. I'm very particular about phrasing and such, so if I changed something, it doesn't mean it was wrong before, it just means I like the new version a tiny bit better, for some reason. If you think I murdered your precious baby, please let me know, I'm sure I won't always get that balance right.

If it's all good, @nicgarner can do his think and copy-edit this.

PR #24, opened 2020-08-26
Multiple Language Support

DrMagPie:
Hey 👋!

So here is multi-Lang Support.
You can see a running version [here](https://drmagpie.github.io)

Out-of-the-box Jekyll with GitHub Pages allows you to leverage a limited, white-listed, set of gems.
Due to that here is implemented GitHub Action, that will build a website on **every** push to `master`, and put it into `gh-pages` branch.

Might be better to switch trigger to some other branch like `release`

Also due to this change, we need to change Branch which is used by GitHub Pages. 
image


LovelySanta:
> Might be better to switch trigger to some other branch like `release`

It seems like master is the perfect branch for this, and rather add a `dev` branch for unreleased stuff

> Also due to this change, we need to change Branch which is used by GitHub Pages.
> image

I assign @ClaudeMetz for this. He can merge in this PR after changing over the released branch.

DrMagPie:
> It seems like master is the perfect branch for this, and rather add a dev branch for unreleased stuff

Well basically I meant not default brunch. as default brunch will receive submission PR's
And `master` usually is default Branch :)

LovelySanta:
> > It seems like master is the perfect branch for this, and rather add a dev branch for unreleased stuff
> 
> Well basically I meant not default brunch. as default brunch will receive submission PR's
> And `master` usually is default Branch :)

I can understand that... In that case we could make a release branch indeed. Can you reflect that in your workflow file then?

DrMagPie:
Let's see what @ClaudeMetz will say on this topic. before switching the branch.

PR #29, opened 2020-08-27
Added pull request template

DrMagPie:


DrMagPie:
@Conor12345 What do you think?
This way this check list will appear in front of us pretty much for every submission :)

Conor12345:
Looks good to me, may need to include SoundCloud example for embedding if anyone decides to use it, but I doubt it will happen so it's good as is for now.

DrMagPie:
Unfortunately it's not as easy to create similar embed for sound cloud... so in the end for user will be easier to add embed manually.

Conor12345:
Okay, fair enough. Looks all good to me then.

PR #38, opened 2020-09-02
Fix translation button being active causing navbar to increase in height

CanadaHonk:
(Not an article submission, styling / SCSS tweak.)


Conor12345 (approved):
Tested in local docker

PR #40, opened 2020-09-02
Add commas between fields in author HTML

CanadaHonk:


CanadaHonk:
Redoing to fix the base.

PR #41, opened 2020-09-02
Add commas between fields in author section

CanadaHonk:


Conor12345:
Tested to be working:
https://media.discordapp.net/attachments/746687101059203103/750698273194377246/unknown.png
https://media.discordapp.net/attachments/746687101059203103/750698511527444582/unknown.png

PR #51, opened 2020-09-03
Better media, Diagrams and Math

### LINKED ISSUE #36: Update submission guide with new media guidelines

DrMagPie:
- Improved media handling, no need to use custom includes
- Added Mermaid for Diagrams
- Added MathJax for math formulas

preview [here](https://drmagpie.github.io/YYYY-MM-DD-ALT-F4-X/)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ group :jekyll_plugins do
   gem "jekyll-paginate"
   gem "jekyll-seo-tag"
   gem "jekyll-polyglot"
+  gem "jekyll-spaceship"
 end
 
 # gem "minima", "~> 2.5"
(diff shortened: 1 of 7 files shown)


Conor12345:
This looks like a much better draft post with some helpful guidance now. Few questions:
Does this change any currently used embeds such as the youtube video in Alt-F4 1?
Does the PR template, readme or discord #instructions need updating with this new formating?
Would it be worth going back and updating previous posts to use the new media embedding standards or would this be a waste of time?

DrMagPie:
1. No it's fully backwards compatible. I did not touch our custom embeds.
1. I believe so. I need to take closer look at those instructions, to say exactly how much stuff needs updating. but as it is backwards compatible, info in PR template readme and #instructions is still accurate. So we just need to add info about additional  functionality.
1. There is no need to update previous posts. Might be only due to consistency reasons, but either way it will work the same way.

Conor12345 (approved):

CanadaHonk (approved):
Looks good.

PR #54, opened 2020-09-03
Designing Blueprints

ClaudeMetz:
### Author info
Name: gamebuster800


ClaudeMetz:
This submission is great, I really like it. Describes pretty much my exact process, in great detail. I asked the author to provide a screenshot or two, and an example blueprint. If he doesn't provide any, maybe someone will have to create those ourselves. It would really improve the submission imo.

PR #149, opened 2020-09-17
Realistic Fusion Weaponry

ClaudeMetz:
### Author info
Name: Romner
Suggested release date:


ClaudeMetz:
This does not have the media handled yet, because it is very big and we need to change it, I think yay is on that already. In the interest in getting this to the translators in time, we'll have to fix this afterwards.

justchen1369:
about the media: I think we need everything to be mp4 for the media to be a reasonable size (<10 MB).

PR #161, opened 2020-09-21
Use new media subdomain for Alt-F4 #5

justchen1369:
One problem is that the thumbnail loads a bit slower now.
Live site: https://goofy-goldwasser-13d710.netlify.app/


justchen1369:
@DrMagPie @Conor12345 can you review?

DrMagPie:
I think that thumbnails can stay in repo.
besides that everything looks good.

also please do `squash and merge` when merging.
image


Side Comment, videos should be size limited to original files size, so it would not stretch Infinity.

DrMagPie (approved):

Conor12345 (approved):
Looks good to me, would it be possible to use some higher res mp4s in alt-f4 #5? Seems like the page size has shrunk somehow by removing imgur from the equation. Presumably this doesnt need to be a vcs change, it can just be changed on the server right?
https://tools.pingdom.com/#5d2d7a6d54800000

Conor12345:
I assume the plan to update all the previous articles with media server links rather the combo of imgur and gh? If so, idm doing it and making sure everything is converted to jpg too

justchen1369:
Yep, I think nukes-new and nukes-old needs some love.

justchen1369:
I think the plan is to use the media server for everything but the thumbnails.

DrMagPie:
> Yep, I think nukes-new and nukes-old needs some love.

Yeah they are stretched and sizing limit should help with it

PR #166, opened 2020-09-23
Translation push

Firerazer:
Chinese translation and corrections for issue 2 and 3

@DrMagPie there is some spaces for the links that I let in the files because it seems I messed up when I edited them last time, hope it will not break stuff that I have no idea ;)


DrMagPie (approved):
All good.
those spaces are within square braces so this is reflected only on formating.
:smile:

Conor12345 (approved):

PR #167, opened 2020-09-23
Chinese Translation - Corrections for Issue3

pigheadx:
Minor corrections and improvement for Chinese issue3

PR #170, opened 2020-09-24
Chinese Translation - Issue 4

pigheadx:
Chinese Translation - Issue 4


Firerazer (approved):
only 1 file added, first commits refer to corrections that are on #167 
Maybe try for next time to have only commits that are relevant for the push ;)

Looks correct for me!

pigheadx:
re-PR with single commit

PR #178, opened 2020-09-26
alphabetize language switcher 

justchen1369:


justchen1369:
@Conor12345 @DrMagPie seems to look ok, can you review?

DrMagPie:
well it looks good. but if we really want to have them sorted at all times I'd better suggest doing something like this
```html
<div class="submenu">
	{% assign languages = site.languages | sort %}
	{% for language in languages %}
	<a {% if language == site.active_lang %} class="button button-green lang-button"{% else %}class="button lang-button"{% endif %}href="{% if language == site.default_lang %}{{ '/..' | append: page.url }}{% else %}{{ site.baseurl }}/{{language }}{{ page.url }}{% endif %}"onclick="setLanguage('{{ language }}')">{{ language }}</a>
	{% endfor %}
</div>

```

And if you want it to have EN as first item of the list then we can do it following way
```html
<div class="submenu">
	<a {% if site.default_lang == site.active_lang %} class="button button-green lang-button"{% else %}class="button lang-button" {% endif %} href="{{ '/..' | append: page.url }}"onclick="setLanguage('{{ site.default_lang }}')">{{ site.default_lang }}</a>
	{% assign languages = site.languages | sort %}
	{% for language in languages %}
	{% unless language == site.default_lang %}
	<a {% if language == site.active_lang %} class="button button-green lang-button"{% else %}class="button lang-button" {% endif %}href="{{ site.baseurl }}/{{ language }}{{ page.url }}"onclick="setLanguage('{{ language }}')">{{ language }}</a>
	{% endunless %}
	{% endfor %}
</div>
```

[Code Reference](https://github.com/AlternativeFFFF/Alt-F4/blob/master/_includes/navbar.html#L18-L25)

justchen1369:
I feel that this is unnecessary as it should always be sorted alphabetically in config.yml.

DrMagPie:
wy to do it manually, when we can automate it? 😄

Conor12345:
Probably is just easier to do it manually since we wont be adding new languages very often

Conor12345 (approved):

PR #179, opened 2020-09-28
Czech translation

Firerazer:
Czech translation by xXmirechekXx
Updated:
- locale
- config
- 1st issue


Conor12345 (changes requested):
- [x] I belive the two letter abbreviation should be `cz` rather than `cs`, I could be wrong though?
- [x] Also, the name of the locale file should just be `locale.yml` not `localeCS.yml`
- [x] The image transcription (and title?) in #1 need translating

Firerazer:
I checked for CS over CZ - I confirm it is CS
https://www.iso.org/obp/ui/#iso:code:3166:CZ (take a look at the translations in different languages + Administrative language(s) alpha-2 ;) )
I take a look and propose corrections for the rest

Conor12345 (approved):

PR #228, opened 2020-10-11
remove unnecessary css rules

justchen1369:


_sass/base.scss
```diff
@@ -79,11 +79,6 @@ mark,
 audio,
 video {
   margin: 0;
-  padding: 0;
-  border: 0;
-  font-size: 100%;
-  // font: inherit;
-  vertical-align: baseline;
```
DrMagPie, line 86: are you sure that these video rules are unnecessary?

justchen1369: It did not seem to affect anything, though further testing is required.

justchen1369:
Closing because I have since learned that it is a part of a reset.css stylesheet.

Issue #10, opened 2020-08-23
Add a site icon


LovelySanta:
Currently the site icon is nothing, defaulting to whatever the browser chooses.
![image](https://user-images.githubusercontent.com/26593477/91635196-bb310500-e9f6-11ea-8eac-2cf1ceb9f643.png)

ClaudeMetz:
Yep, I'll contact the person working on this soon to see how it's going. Don't want to time-pressure them too much yet.

ClaudeMetz:
Can someone try and set the favicon on the website to the new logo?

Conor12345:
![image](https://user-images.githubusercontent.com/34628734/91765554-19194480-ebd1-11ea-9636-806574d69864.png)
Picture during testing

PR #238, opened 2020-10-16
Nauvis Archives: Jargon Update

heinwessels:
### Author info
Name: stringweasel
Suggested release date: #10 (as an update on #9)

This is an update on the name "Nauvis" after Klonan shared some interesting information on Discord.


submissions/nauvis_archives_jargon_update/nauvis_archives_jargon_update.md
```diff
@@ -0,0 +1,7 @@
+### Nauvis Archives Update: Nauvis Backstory
+
+In [Alt-F4 #9](https://alt-f4.blog/ALTF4-9/) we investigated the name of our planet, Nauvis. Especially based on the fact that this name is never mentioned in the game (well, almost, thanks [IronCartographer](https://discordapp.com/channels/139677590393716737/603392474458882065/766621041454874626)). However, soon afterwards, the Wube Community Manager, [u/Klonan](https://www.reddit.com/user/Klonan/), [graced](https://discordapp.com/channels/139677590393716737/603392474458882065/766687476198539264) us with information never heard before.
+
+![](media/klonan_on_nauvis.png)
```
Conor12345: This will need updating to be a quote like in PR #248

Firerazer:
are the links to specific discord messages pertinent (ironcartographer + klonan message links)? should we use the new quote function and copy/paste in this the message content?

Conor12345:
> are the links to specific discord messages pertinent? should we use the new quote function and copy/paste in this the message content?

My plan was to use the new quote styling but those aren't quite ready yet so I'll add them tomorrow or Friday probably. I'm not sure about including the discord message links, its doesn't hurt but idk if two links are necessary, maybe just one instead?

heinwessels:
> are the links to specific discord messages pertinent (ironcartographer + klonan message links)? should we use the new quote function and copy/paste in this the message content?

It's not pertinent at all. But I don't want to include the actual quotes since it will break up the text too much. We can rather remove the discord links. 

>My plan was to use the new quote styling but those aren't quite ready yet so I'll add them tomorrow or Friday probably. I'm not sure about including the discord message links, its doesn't hurt but idk if two links are necessary, maybe just one instead?

There no difference between one and two links. It should be either add both or remove both. But we might be getting very pedantic on something very minor. I don't mind the links being there, but you can remove both if you want. As long as the flow of the text does not change.

ClaudeMetz:
I changed quite a bit about this submission, since I didn't feel like it really said what I wanted it to say. Hope you're okay with that weasel. I also took out the speculation about the reasons for the scrapping at the end, don't think that's something we should be doing in this case.

submissions/nauvis_archives_jargon_update/nauvis_archives_jargon_update.md
```diff
@@ -0,0 +1,20 @@
...
+
+![A screenshot of the alert message mentioning Nauvis](https://media.alt-f4.blog/ALTF4/10/nauvis_mention.jpg)
+
+After this first was made in the Discord, Wube Community Manager Klonan dug up some never seen before [backstory](https://discordapp.com/channels/139677590393716737/603392474458882065/766687476198539264) for us to enjoy:
+
+{% capture quote_text %}
+This is all you get:
+```The cova system is named or the orange dwarf star at its centre. Four known planets orbit the star, as well ass numerous moons, unslisted dwarf planets and asteroids. Closest to the star is Nauvis - this is the planet upon which the game takes place. A tempered world capable of sustaining life.```
+And maybe this... <br/>
+```The player, named Alden Winters...```
```
Firerazer, line 16: the ``` formating seems to be broken there

heinwessels: I'm looking at it ATM. I'll push soon.

Firerazer: check #248 - it's the multiline ``` formating that is broken. will edit later if not done since ;)

heinwessels:
@ClaudeMetz that's no problem, And I was on the fence of my speculation since it is dangerous territory. I will just go through it and tweak it slightly, maybe different word choices, just to make it feel more like my work again. Nothing big. You can just redact my commit if it's too close too publish and you don't want to go through it again.

One thing though, here:

>  For one, it is used internally in the code to designate the main surface of the game. We were well aware of this fact, but decided not to include it as a normal player will not run into it in that way.

I explicitly mentioned this in the original article. Here:

>Well, the game itself has various dimensions of existence—or as the code calls them—surfaces. And the surface we all live on is called nauvis.

What we missed it that it's displayed in some cases while using scenarios or the editor. I'm changing the piece to reflect that.

I also fixed the quote, and a few other little things. But it was good editing, thanks Theranas.

Conor12345 (approved):

PR #246, opened 2020-10-19
wrapped countdown in paragraph tags

### LINKED ISSUE #193: Timer layout issues
Currently, the timer we sometimes embed behaves weirdly. If you just put it in, it overlaps with the bottom text. The quick solution was to add a linebreak after it, but that makes it have a big margin on bottom and a smallish one on top. It would be great if we could find a way to have it have the same padding/margin on top and bottom.

DrMagPie:
closes #193
[Preview](https://drmagpie.alt-f4.blog/ALTF4-3/#event-molten-silicon-jd-plays)


Conor12345 (approved):
Looks good, just need to add the new format to the draft file

PR #261, opened 2020-10-24
A scientific examination of the Factorio world

pocarski:
### Author info
Name: 
Suggested release date: 

- [ ] Images are hosted on Media server 
- [ ] Gifs are hosted on Media server
  - [ ] Gifs are [compressed](https://www.iloveimg.com/compress-image/compress-gif) 
- [ ] Videos are hosted on Youtube or Media Server
  - [ ] Make sure video is public or unlisted
  - [ ] Use provided Youtube embed ``
- [ ] Audio is hosted on souldcloud
  - [ ] Embed soundcloud post of audio file


Conor12345 (approved):
Ready for release

heinwessels:
I have experience in technical writing and I have a few suggestions on how to make it more readable.

- A big assumption is that the planet is spherical. This needs to be mentioned.
> Assuming a spherical planet, and using the equations for centripetal acceleration and gravitational acceleration, we can show that:

- Add the second `m` in the first line, otherwise, you skip steps to early.
> $ \frac{GMm}{r^2} > mrw^2 $

- After the formula wall you need to specify what all the symbols mean. Or, even better, every time after you introduce a new symbol, you declare what it is. This is typically done like this right below the equation:
> where $/rho$ is density, $V$ is volume, ...

- The comparison densities should be in the same unit as the calculation, like in the original reddit post. Otherwise, it's not easy to compare.

- Adding `the mass of the sun` can confuse people since the sun is a star. I think it already sounds majestic just mentioning `a star dies`. If you want to leave it in, rather say `the mass of our sun`.
>left after a star dies

- Maybe specify what Nauvis is probably tidally locked to.
>tidally locked with it's star

- Is there a link to the original reddit post (or comment) talking about orbital mechanics? Or the part about Nauvis' albedo? Or nukes being underpowered?

- Inconsistent use of °C throughout the article. Sometimes it's written degrees C, sometimes not. It should be consistent.

- For the calculated size of the explosion, 60 tiles doesn't sound much. How big is the explosion currently? How much bigger is the calculated than the actual?

- In the formula $N=mav/600$. Don't skip the first step then, rather write
> N = mav/P = mav/600..

- Explaining the relationship between efficiency isn't very clear. It could be:
> So, since fuel efficiency divided by ... new acceleration, this means that nuclear fuel... 

- You don't need so many decimal places for efficiencies, it makes it very hard to read. Typically you only use 2 decimal places.

PR #265, opened 2020-10-27
Optimize background images

justchen1369:
If you could add the hacktoberfest labels that would be nice too.


Conor12345 (approved):

Conor12345:
Can be merged after @DrMagPie reviews

DrMagPie (approved):
Looks good. but if we are removing `webp` shouldn't we get rid of that probing script?

DrMagPie:
this one https://github.com/AlternativeFFFF/Alt-F4/blob/dcd395da387639c3610f9a653029d60602e30ec4/assets/GLOBAL/js/modernizr-webp.js#L3

Issue #259, opened 2020-10-24
Add superscript + subscript


Conor12345:
This will complete the basic markdown support for the site.
- [x] Superscript
- [x] Subscript
- [ ] Strikethrough tweak if possible?
```
# Heading 1
## Heading 2
### Heading 3

**Bold**

*Italics*

***Bold and Italics***

~~Strikethrough~~

superscript<sup>abc123</sup>

subscript<sub>abc123</sub>
```

![image](https://user-images.githubusercontent.com/34628734/97084448-09d1d880-160f-11eb-9fa7-9e87908671c0.png)

PR #269, opened 2020-10-30
Update to strikethrough style

jmannfeld:
### Author info
Name: Jordan Mannfeld
Details: Hi! I was looking into the options for editing the default styles of strikethroughs in Markdown - #259 . Unfortunately there is not much out there in the css spec for stying the strikethroughs using the `~~` syntax. Instead it is recommended to use the `<s>` html element to strikethrough text. This also allows for easier styling. I didn't change a whole lot, but showed how you are able to tweak the styles a bit. I added a `text-decoration-color: red` for the strikethrough color because it's probably the most applicable option. Other possible styles include:

- text-decoration-style (solid, wavy, or dash)
- text-decoration-thickness (I would have used this property, but it is hardly supported in any browsers right now)

Screen Shot 2020-10-30 at 10 42 51 AM

### DIFF
--- a/_sass/base.scss
+++ b/_sass/base.scss
@@ -146,3 +146,7 @@ sub {
   vertical-align: sub;
   font-size: 0.7rem;
 }
+
+s {
+  text-decoration-color: red;
+}


Conor12345:
image
This doesn't really fix the issue of the height of strikethrough line and it being red adds no value, if anything it makes it worse.

jmannfeld:
Okay, the issue just said "Strikethrough tweak" which is pretty vague. I'd still recommend using the `<s>` element if you want to style your strikethroughs in the future.

DrMagPie:
first of all. @jmannfeld Thank you for trying to help! it's really appreciated. 
And yes the issue is a bit vague. Perhaps we should be more explicit in case someone "external" wants to contribute. (So far we were discussing it in details on discord)

in either case, as Conor mentioned issue is that line is a bit shifted up.
`~~` Markdown syntax is translated to `<del>` element 
image

which also can be styled. so applying styles also is not an issue.
Unfortunately, I didn't find any means of changing lines vertical positions. so I assume that currently, it is impossible. (Would love to be wrong.)

 @Conor12345 I think that this is caused by a font. we should play around with it.

justchen1369:
Kind of the same issue with discord and its font as well I think.

PR #277, opened 2020-11-03
Complexity Corner - Productivity modules

Conor12345:
### Author info
Name: TheKool (needs checking with author)
Suggested release date: 

- [X] Images are hosted on Media server 
- [X] Videos are hosted on Youtube or Media Server


dexteritas:
In contrast to the previous articles, a double space is used here after each sentence. I think this should be done uniformly.

ClaudeMetz:
> In contrast to the previous articles, a double space is used here after each sentence. I think this should be done uniformly.

I think markdown ignores the double spacing automatically, but I removed them anyways.

PR #280, opened 2020-11-05
Nauvis Archives: One Rocket Defense Per Minute

heinwessels:
### Author info
Name: stringweasel
Suggested release date: 

This piece still needs work on the media, which I should finish before Monday. I also share it with DaveMcW to make sure I got everything right, so if he has feedback it might change slightly. 

I might have gotten a little carried away with the length though. So I created the PR so long since @ClaudeMetz and @Conor12345 mentioned that they want shorter pieces. If the length is an issue then I could try to trim it somewhere or slash it if it's really necessary. But then I would need some feedback on which parts you guys like more so I can trim the other part :P

Next time it will be a nice short and compact piece again.


Conor12345:
We don't necessarily want shorter pieces, we just want to shorten the length on the releases. This means we'll probably do longer submissions on their own as a full release rather than pairing it with another submission. Also means shorter submissions can be paired together to make a normalish length release. Basically, make it as long or short as you want and we'll work around it.

Conor12345:
Firstly, this is an amazing submission!! Reading through, there are a couple places where I wanted more information but it seems like you'll do this with the extra media (normal pump vs small pump is one as idk what a small pump is). 

I also think it doesnt quite convey the scale of the base, which is admittedly very hard to do on this kinda scale. I was thinking maybe we could embed the http://twinsen.info/Web/Map(197,205)1024/ map at the end of the piece to allowing people to natively look around the base without having to leave the site? I'm pretty sure its technically possible but just wanna check if thats something you'd want or not before I figure out the exact setup for it?

I will fix all the embeds after editing in the shiny new ***technical review*** phase and also I can upload the mod file and save file to the server at that time too.

AMAZING WORK - KEEP IT UP!!!!!!!!!!!

Conor12345:
Initial testing of the interactive map embed thing. 
image
The very, **very** janky html for it. Will need doing much more properly but I can handle that if you like the idea :D
```<div class="mapouter"><div class="gmap_canvas"><iframe width="1000" height="700" id="gmap_canvas" src="http://twinsen.info/Web/Map(197,205)1024&t=&z=13&ie=UTF8&iwloc=&output=embed" frameborder="0" scrolling="no" marginheight="0" marginwidth="0"></iframe></div><style>.mapouter{position:relative;text-align:right;height:700px;width:1000px;}.gmap_canvas {overflow:hidden;background:none!important;height:700px;width:1000px;}</style></div>```

heinwessels:
That's something I struggled with, is to convey the crazy scale of this base. So I really like the interactive map idea! I think if it works smoothly we can definitely give it a go! And if there's something else that you think I could explain better or show with a picture, like the small pump, please tell me. Most readers will be like you and not know how it was 5 years ago, so any input will really be great!

And thanks! I really appreciate your excitement :D Glad I could get at least one person excited about this base so far :P

Conor12345:
Awesome. I think showing the new pump and old pump in the same setup using the [compare embed](https://alt-f4.blog/ALTF4-X/#compare-images) thing (the one with the slider) should make it clear. May be worth adding another sentence around that section if you can think of something that fits nicely, if not its probably fine. Seems like a weird old mechanics and those are definitely interesting, so drawing more attention to it with a small tangent would definitely be cool.

I'll get to work on the map embed thing over the next few days

Nanogamer7:
**Fact Checking:**
> the record was likely held by u/AccidentalChef with his vanilla-ish 10rpm base [line 3]

While not necessarily wrong, [this](https://www.reddit.com/r/factorio/comments/655440/) semi-vanilla base by u/Stevetrov managed 15.7 rpm. I'd consider it to be the record, as it is still in the same version (0.14), but it was made on a later date.



> full-productivity-bonus 2kspm factory [line 23]

closer to [2.5kspm](https://kirkmcdonald.github.io/calc.html#zip=fZJJbsMwDEVv41UEOAm6MaDDMBTjEtUEkup0+spwN03lbAiI7+tzkAIY+LOb3Txh9dcpce7xRtE8fVYhVWcCWWsRc1t6CsnX6xRuXreI/nyZ2Ciph2YlgXHJTpEpI7kK+LbIcnmZ51MsK6sxDmHiyAbyNYT4SokR4hBWKaHhYdVmm/PYV/uBRmRK3TNSHwmxpRbBiiy5xfgQThQJTfpM9ya5mw1FdxYCSS7BCt+cx6JVKEMYM85KYjRuoTIJcl6dlJaDPq9SuR6A3y2+b7vapx/qBAKM+xDgx9wOFIxif4OxoZYI0nef6eB2JQrPOlIrmdwH/CM7/vt5R5If) with 80k green circuits/minute ([1rdpm raw ingredients](https://kirkmcdonald.github.io/calc.html#zip=jY/BCsMgEET/xlMEk9BLYD/G6FKWqhFdQ/v3NZhLU6G9DOzsmx3WatYwSiWV8BRgFis6BnzGhDlLTjrkuCWWhy2shzgLu0I+1MA4CWL0GXJEtNJvtjiU85KWmxpiqqNh2olfl5W2uw6mJgwlU4irO7aEqa0U7rKE5io1VPOB5yDana/CUJy7yPCBTD+RLtD/4V+yX9ohu9z5d2f1Bg==))

Every other 'fact' I checked seems to be correct, good job ;)

---
Other things:
- [line 11/15] DaveMcW is mentioned 2 times right after one another, with two different link. I think this isn't the best solution for this, but I can't think of a better one either
- [line 25/29] above and below `MEDIA (media/base_old_circuits.mp4)` there is two times the same phrase *"Notice the amount of inserters"* & *"notice how the chemical plants producing plastic have four output inserters"*
- [line 33] Maybe change the link to chain signals to https://wiki.factorio.com/Rail_chain_signal#History
- [line 61] Two sentences starting with *"For example"*
- [line 61] *"50 times more powerful."* I feel like the link, while funny, is a bit confusing, as at least I expected something on how this was changed and not a meme
- [line 74] The reddit comment has some tracking stuff on it (`?utm_source=share&utm_medium=web2x`), for privacy reasons I'd like to recommend to remove such stuff from reddit links

Issue #120, opened 2020-09-11
The `a` (link) element for clicking on the thumbnail image taking to the post possibly interferes with screen readers


CanadaHonk:

CanadaHonk:
Unknown whether this is a full / proper accessibility issue - this seems to have been a well-intended feature for HTML5 so it's possible screen readers are fine with such elements. Closing for now.

PR #288, opened 2020-11-07
Whole card is a link

### LINKED ISSUE #285: I expected the whole post box, in the index, to be a link
I expected that tapping anywhere in the box for each index entry would take me to the content, not just the title link. 

Given it is the most likely thing I want to do, making that active across the whole box, and then allowing links in the index entry (eg: author) to take me to more specialized places, is the least surprising choice.

DrMagPie:
Closes #285
[Preview](https://drmagpie.alt-f4.blog)

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -6,17 +6,15 @@ layout: default
 	{% include pagination.html %}
 	{% for post in paginator.posts %}
 	{% unless post.draft %}
-	<div class="panel-inset-lighter blog-card">
+	<a href="{{ site.baseurl }}{{ post.url }}" class="panel-inset-lighter blog-card">
 		{% if post.image %}
 		<div class="blog-card-thumbnail">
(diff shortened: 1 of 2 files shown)


Conor12345:
Is this likely to cause accessibility issues with screen readers? I know @Oj18 has issue #120 open but I'm not sure how the two interact.

Conor12345 (approved):

justchen1369:
Tbh I don’t think that this is beneficial, seeing as now there’s no feedback about the card being clickable.

Conor12345:
Ye thats a good point, didn't notice that actually

justchen1369:
Another thing is that, imo, it doesn't seem like the card should be clickable- just the title.

Conor12345:
I quite like that the whole card is clickable, but it would be nice if you could still hover over the title and have it be highlighted. Especially good for mobile too

PR #290, opened 2020-11-10
Update spidertron (performance fixes)

xthexder:
In an effort to improve the performance on mobile, I've implemented `u/gtsteel`'s suggestion to use `transform:` instead of `left:`.
Removing `width:` and `height:` changes and using scale transforms instead will also improve performance, but will require a more complicated change, so that's for a later PR.

I've also added some code to the javascript to prevent Spidertron from causing a page resize when walking off the bottom or right of the screen, which causes a ton of lag on mobile.

### DIFF
--- a/_sass/base.scss
+++ b/_sass/base.scss
@@ -14,7 +14,6 @@ body {
   font-size: 100%;
   display: flex;
   flex-direction: column;
-  height: 100%;
 }
 
 body {
--- a/assets/GLOBAL/css/spidertron.css
+++ b/assets/GLOBAL/css/spidertron.css
@@ -39,21 +39,21 @@ SOFTWARE.
 }
 
 .spidertron-home {
-    width: 1px;
-    height: 1px;
+    width: 0px;
+    height: 0px;
 }
 
 .spidertron {
     position: absolute;
-    top: var(--spidertron-location-y);
-    left: var(--spidertron-location-x);
+    top: 0px;
+    left: 0px;
     contain: strict;
 
     width: var(--spidertron-width);
     height: var(--spidertron-height);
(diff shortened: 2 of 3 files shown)


Conor12345:
Is there a live test site available?

Conor12345:
I have updated test.alt-f4.blog with the performance tweaks

justchen1369 (approved):
I'm not sure that it helped very much, but it does work. :)

xthexder:
I didn't see much of a performance change unless I also removed the width/height animations, so I'm not expecting much there.
The bigger fix is for lag when spidertron walks off the screen on mobile. The page shouldn't change size anymore, and be a pretty consistent frame rate.

DrMagPie:
https://media.alt-f4.blog/fileshare/weirdbd.mov

Browser: Latest version of Firefox Developer edition.
For some reason Background flashes/glitches....
Issue is present not only when spider is activated.

justchen1369:
I've seem that occur when the site is still loading- can you see if it happens on the normal site?

justchen1369:
That's a yes.

DrMagPie:
hmmm,

Have looked again and now it seams not to happen anymore...

but first time page was loaded fully when this was happening.
It looks like some styles were changed on the fly due to what background was replaced/resized.
Might be due to cache? although for test subdomain it should be disabled... hmmm

justchen1369:
I think this happens because sometimes the site is longer than the background- fix is probably the make the text "floating" above the background.

Conor12345:
Is this ready to be merged?

xthexder:
It seems like any issues are unrelated to this PR, so it can go in any time.

PR #291, opened 2020-11-11
Credit tweaks + layout updates

Conor12345:
Layout tweaks to more prettify the site whilst allowing more editors
image
image
image


DrMagPie (approved):
I have changed it a bit. Just to use conditions in author.html and to reduce duplicate code :)

Issue #294, opened 2020-11-13
Footer buttons look greyed out/unavailable


Nanogamer7:
Currently the buttons in the footer (about, github, discord, and atom feed) are the same color as the social media buttons on the factorio.com website, which isn't ideal by itself, but especially where it is placed it looks to me like an empty space. Furthermore the buttons have the same (or at least a similar) color to greyed out buttons for the next page, so even if you do spot the dark grey buttons between the rest of the dark grey UI, it might look like the buttons are not functional

![Footer buttons compared to next issue button](https://user-images.githubusercontent.com/45211068/99074923-eae4b780-25b8-11eb-83c4-5893e5c435cd.png)
*The dark grey footer buttons have the same color as the "next issue" button, which is unavailable, since I'm on the last issue*

My suggestion now would be to either make the buttons the same color as clickable buttons above, or, even simpler, make the logos on the buttons brighter. This should be relatively minor change to make the contrast higher and thus the UI/UX better; but if you don't agree constuctive critisism/feedback on my suggestion would be appreciated ;)

DrMagPie:
<img width="542" alt="image" src="https://user-images.githubusercontent.com/23652900/99322375-7b9ceb00-2878-11eb-9c87-d56331c0b44c.png">


I'm not sure about these looks.
footer should be subtle, but with these bright buttons it poops out to much (In my opinion)

Nanogamer7:
Gotta agree with the "subtle", and that this definitely doesn't fullfill the requirement, **but** I think now it at least looks like there is something there. I don't have an idea right now, but I'll try to come up with something that solves both issues ;)

DrMagPie:
FYI you can see it live here https://drmagpie.alt-f4.blog

Conor12345:
Having played around with the styling I think this is the best possible look. They stand out much more which I have no issues with as they are underused.

PR #295, opened 2020-11-13
initial buttons.scss refactoring

justchen1369:


Conor12345:
Is there any particular reason for this refactoring? Also could we do anything about #294 here?

DrMagPie:
@Conor12345 nope, not related to #294 
Here he just removes duplicate code and uses SCSS magic to import it from one generic class.

@justchen1369 I think it's better to leave `.active` incase we want to use it in future.

justchen1369:
Doesn't `:active` serve the purpose of `.active`?

DrMagPie:
yes that true. but there could be the case when it can't be determined on a render stage, then `.active` comes into play.
in theory... XD

justchen1369:
Someone could just reference this PR for the correct additions :D

justchen1369:
If there are no issues then I'll just merge this.

DrMagPie (approved):

PR #296, opened 2020-11-15
add highres include

justchen1369:


DrMagPie (commented):
I would name parameters a bit differently. so they would better represent what they are doing. like (`href` for link to original and `src` for image it self. but this is minor.)

with regards to include itself, have you tested it? I feel like it's missing some css classes. (like `media`)

justchen1369:
I was unsure, so that's why i did the PR.

Conor12345:
Would it be possible to merge some of our includes? Currently we have a bunch of different ones for different scenarios which is real annoying. We will need to keep old ones for compatibility reasons but it would be nice to have two smarter includes (one for video, one for image) that have all the functionality included (caption, highres etc.)

justchen1369:
Either that or document all of the includes somehow?

justchen1369:
I was also thinking of moving all of the site includes to their own folder.

Conor12345:
Being resolved with PR #305

PR #297, opened 2020-11-15
Moddability - The Birth of a Mod

DedlySpyder:
### Author info
Name: DedlySpyder
Suggested release date:


heinwessels:
One last bit of feedback is the paragraph at line 11 could be a little clearer. I think it could benefit from having an extra sentence, or different wording, to ensure that you get your message across. But other than that it's a great submission, it's an interesting read and well written!

Nanogamer7:
**Fact Checking:** 
- ❗ [[WIKI](https://wiki.factorio.com/Tutorial:Modding_tutorial/Gangsir#Load_order)] Mods usually add new items, entities, etc. on the **second** pass, the "data stage".

---

Further notes:
- The credit to the xkcd should probably go in a new line, not entirely sure though
- You left the last image unexplained, and while it doesn't really require that much or even any explanation, I personally think it would look better with at least a short caption underneath ;)

---

IG = In-Game
WIKI = Wiki (duh), expect the specific page to be linked for clarity
LO = Logic, article doesn't agree with itself
MP = Mod Page
MA = Mod Author
F = Forums
FFF = *self explanatory*
RD = Reddit
N/A = Nothing saying otherwise, thus no reason for it to be different (search term linked)
YT = YouTube video/comment

DedlySpyder:
> * ❗ [[WIKI](https://wiki.factorio.com/Tutorial:Modding_tutorial/Gangsir#Load_order)] Mods usually add new items, entities, etc. on the **second** pass, the "data stage".

I wasn't referring to settings vs data stages, I was referring to how the mods are loaded 3 times by `settings/data`, `*-updates`, and `*-updates-final-fixes`. The [wiki](https://wiki.factorio.com/Tutorial:Modding_tutorial/Gangsir#The_data_stage) recommends:

> it is recommended to create prototypes as early as possible

If you think the data/settings distinction should be explained better I can reword that paragraph.

Nanogamer7:
That's still ambigous then, and I don't think technical articles should leave much/any room for interpretation, so if possible reword it a little please ;) (Although I don't think this is entirely on your side, as I think the wiki article nests its use of "phases" a bit to much.) A suggestion would be "The wiki explains this in more detail, but in short most mods add their new content right after finishing the settings phase, which allows for mods that want to implicitly rely on other mods to do it without needing to even know they exist."

DedlySpyder:
Ok, I rewrote the paragraph that talked about the data stage and cleaned up the credits. Plus added a few words elsewhere.

nicgarner (approved):

PR #305, opened 2020-11-16
Generic image include

DrMagPie:
[preview](https://drmagpie.alt-f4.blog/ALTF4-10/)

combines:
- [X] figure.html
- [X] hires.html
- [X] imgur.html (kinda does we just need to use full link)
- Something missing?

TODO:
- [ ] Update Sample post

### DIFF
--- a/_includes/image.html
+++ b/_includes/image.html
@@ -1,6 +1,6 @@
 <p></p> <!-- Spacing -->
 <div class="image">
-  {% if include.href %}<a href="{{ include.href }}">{% endif %}
+  {% if include.href %}<a href="{{ include.href }}" target="_blank">{% endif %}
     {% if include.caption %}<figure>{% endif %}
       <img src="{{ include.src }}" alt="{{ include.alt }}">
       {% if include.caption %}<figcaption>{{ include.caption }}</figcaption>{% endif %}
--- a/_includes/image.html
+++ b/_includes/image.html
@@ -1,0 +1,11 @@
+<p></p> <!-- Spacing -->
+<div class="image">
+  {% if include.href %}<a href="{{ include.href }}">{% endif %}
+    {% if include.caption %}<figure>{% endif %}
+      <img src="{{ include.src }}" alt="{{ include.alt }}">
+      {% if include.caption %}<figcaption>{{ include.caption }}</figcaption>{% endif %}
+    {% if include.caption %}</figure>{% endif %}
+    {% if include.href %}<i>Click to view full resolution</i>{% endif %}
+    {% if include.href %}</a>{% endif %}
+</div>
+<p></p> <!-- Spacing -->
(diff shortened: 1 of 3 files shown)


justchen1369:
I can spend some time changing all of the old links to be imgur links.

_includes/image.html
```diff
@@ -0,0 +1,11 @@
+<p></p> <!-- Spacing -->
+<div class="image">
+  {% if include.href %}<a href="{{ include.href }}">{% endif %}
```
justchen1369: target should be _blank for the link.

DrMagPie: what do you mean? why?

Conor12345:
Not sure I like the way the click to show high resolution shows up on top of the image. Would it be possible to have it below instead?

DrMagPie:
Actually it's easier to align it to the top cuz caption on bottom can wrap
image
image


also this is an option
image
image
but on mobile it looks like shit

_sass/image.scss
```diff
@@ -0,0 +1,28 @@
+.image {
+  position: relative;
+  text-align: center;
+  a {
+    text-decoration: none;
+    i {
+      position: absolute;
+      top: 50%;
```
DrMagPie, line 8: @Conor12345 Fell free to play around with this parameters to align it the way you want.
`top:` and `bottom:`
ref: https://www.w3schools.com/howto/howto_css_image_text.asp

_includes/image.html
justchen1369: This way clicking on the link will open it in a new tab.

Conor12345: Clicking needs to open a different, higher res version so the link needs to be explicitly given.

DrMagPie: mm makes sense

DrMagPie: @Conor12345 yes. `href` is high res version and `src` is version displayed on website

DrMagPie:
```
{% 
include image.html 
href='https://media.alt-f4.blog/ALTF4/13/base_new_map_og.jpg'  <!-- hires/original -->
src='https://media.alt-f4.blog/ALTF4/13/base_new_map.jpg' <!-- optimised -->
alt='Map view of rail network, showing only an eighth of the outposts.'
caption='Map view of rail network, showing just 27 of the 216 outposts.' 
%}
```

Conor12345:
Could we permanently update the draft page to include these too?

Conor12345:
Also is it possible for the image to open in a new tab, rather than the current one? Unless there is some reason it shouldn't?

justchen1369:
That's what target _blank is for.

Conor12345:
> That's what target _blank is for.

Yep my bad, was on test site which hadn't been updated.

PR #336, opened 2020-12-06
Optimising CountDown include

DrMagPie:
pre-calculating time values to reduce calculations/operations per loop iteration


justchen1369 (approved):
LGTM

PR #338, opened 2020-12-06
excerpt first paragraph

DrMagPie:


Conor12345:
Would it be possible to have the embed only include the first paragraph too? 
image
Im guessing it would have to be added here with the image thing we added.

DrMagPie:
Where that thing gets text from?
if from description then it already should have what we need
image

PR #344, opened 2020-12-10
Centered countdown embed

DrMagPie:


Conor12345:
Think the padding needs to be changed as the countdown is overlapping the text now in old releases. See Alt-f4 3,7 + 12 for examples
image

PR #345, opened 2020-12-10
Vertical Quote

DrMagPie:
[Preview](https://drmagpie.alt-f4.blog/ALTF4-1/)

```j2
{% include quote.html vertical='true'  text='Text' translation='Translation' author='Conor_' %} // vertical quote
{% include quote.html text='Text' translation='Translation' author='Conor_' %} // regular / current version
```


Conor12345:
@DrMagPie Would it be possible to have the author always below the text? Feels a bit weird having three columns.

DrMagPie:
@Conor12345 what do you think about this solution (same link above)?

Conor12345:
> @Conor12345 what do you think about this solution (same link above)?

That looks perfect thanks 👍

PR #355, opened 2020-12-28
Issue 19 - Retrospective

ClaudeMetz:
This is a very rough draft - subject to change if we split it up into two issues. Not yet sure where to put the conclusion, probably makes more sense in the second issue


Nanogamer7:
Haven't had too much time for fact checking during the holidays, but due to the nature of this edition, there shouldn't be to much that could be wrong, and I couldn't spot any obvious errors during proofreading of the translation ;)

PR #361, opened 2021-01-05
 DE #19 corrections 

dexteritas:


Firerazer (changes requested):
Based on discord comment, the " is a formating thing, hence have to be handled by the language translators. Please adapt your text to it, and eventually review all others submissions

Rest is approved

dexteritas:
Since the quotation marks are otherwise also replaced by two separate characters (as is correct in English), it should be adjustable. At least I know that it works automatically on other websites, and you can set the language.

I have adjusted it for this now anyway again manually, in the previous issues it is already so.

Firerazer (approved):

Conor12345 (approved):

PR #380, opened 2021-01-17
Create Complexity Corner 3 - Efficiency Modules

jmkool:
### Author info
Name: The Kool


submissions/Complexity Corner 3 - Efficiency Modules/Complexity Corner 3 - Efficiency Modules.md
```diff
@@ -0,0 +1,35 @@
...
+
+Even if you do generate all your power from solar, efficiency modules will still reduce pollution made by the machine itself, because that pollution is modified by the power usage multiplier.  In simpler words, this means that if your machine is only using 20% power because of efficiency modules, it is also producing only 20% pollution.  In this example, an Assembly Machine 2 would only produce 0.6 pollution per minute, down from the normal value of 3 per minute.
+
+To show how drastic an effect efficiency modules can have on your pollution cloud, I've prepared a factory that is roughly representative of an average player's pre-rocket factory, maybe even smaller.  It makes 45 SPM, makes all its power from solid fuel and steam, and uses mid-tier machines (steel furnaces, electric mining drills, Mk2 assemblers).  I ran this factory three times, starting each run with no pollution.  The first run was with no modules, the second run used a mix of productivity and speed 1 modules, and the third run used efficiency 1 modules everywhere.  Take a look at the pollution clouds after 1 hour.
```
heinwessels: Well done one the thorough and hard work! I think you should just be concise about _why_ you started the factory from scratch each time. Essentially because you want to include pollution created by making the modules.

Nanogamer7:
**Fact Checking:**
- ✔ [[Factorio-Wiki](https://wiki.factorio.com/Module#Efficiency_module)] Efficiency modules reduce power & pollution by 30%|40%|50% and cannot go below 20%
- ❔ While it isn't wrong, that *some* machines produce pollution on their own, I wouldn't necessarily use *some* for *(nearly) all crafting machines*
- ✔ An assembling machine 2 with 80% less pollution produces 0.6 instead of 3 pollution per minute
- ❗ Nest don't, or shouldn't send any attacks if they don't absorb any pollution, assuming no enemies are within range of the biters themself

---

Further notes:
- You might already be adding this with you images, but an explanation for you test results would be welcome 
- Expect a small update once you added images ;)

---

✔ Correct
❗ Incorrect
Facts and Statement without any symbols weren't validated by me; that doesn't mean they are false, it could be something personal (closed-source, etc.) I can't check. If you feel like a claim I left out should be validated, please tell me!

Conor12345:
- [ ] Needs media adding
- [ ] Need techedit

PR #382, opened 2021-01-21
Nauvis Archives: Issue #9 Editors Note

heinwessels:
We wanted to add an editors note to Issue 9 about the new information that someone gave, and the cool new info that Klonan gave. It just never happened and has been sitting on my mind. We just need the translations for it as well.


Firerazer:
- [x] FR
- [ ] ES
- [ ] IT
- [x] DE
- [x] ZH

Firerazer:
ES and IT are not active, I would consider this ready for being pushed.
@Conor12345 do we maybe want to edit the ES and IT piece with the english text?

PR #389, opened 2021-01-23
Added possibility to name languages

### LINKED ISSUE #373: Language select to narrow for PT-BR
In the language drop down, the gap on the right side of portuguese is too narrow compared to the rest of the UI
Screenshot 2021-01-14 at 17 05 35

DrMagPie:
Addresses #373 
image
[ref](https://discord.com/channels/745990677606826005/746687101059203103/799950070342484028) For internals only :)


Conor12345:
They call me the bug find magician:
On non-widescreen monitors is extends the page weirdly, should probably be aligned to the other side instead?
Also the font weirdness for non-standard character in RU and ZH - Maybe finally time to deal with the font issues we've had in the past and self-host fonts and stop that weird text moving glitch and the slow loading of fonts in general/
image

Same issue is also present on mobile with similar weirdness
image

Probably want to do a more full testing 
Also im getting all the language strings so we should have full data for testing soontm
This is for sure gonna be an improvement so good work :D

PR #391, opened 2021-01-23
Logistic Belts

Recon419A:
Add a submission on the topic of logistics belts, along with supporting media.

### Author info
Name: Recon419A
Suggested release date: Any


submissions/logistics_belts/logistics-belts.md
```diff
@@ -0,0 +1,27 @@
...
...similar concept: sushi belts. Like logistics belts, sushi belts place more than two resources on a belt, but they generally do so as a loop designed for complicated crafting. A circuit condition at the entrance or around the entire belt helps to control the amount of every resource present, even leading to such complicated examples as the "bloodbus" - a circulatory system for your factory (https://imgur.com/gallery/Q4oR0#Jiwed0Q). While the concept of a logistics belt is similar, it has a few important differences, and the key one is related to their function: while sushi belts go *around*, and tend to service multiple assemblers in no particular order, logistics belts go *to*, and tend to direct resources from one location to another. In fact, a logistics belt can be used to feed a sushi belt in place of an ordinary bus - when resources are needed, they can be sent, with carefully limited rate control preventing over-population of the belt loop. Other concepts like clocked belts or inserter limiting can be applied here - and can be combined with logistics belts to achieve monumental feats of engineering. Logistics belts themselves are just one part of the growing player's toolkit.
```
Conor12345: Here could we embed a picture of the blood bus factory or just a regular sushi belt for anyone that hasn't seen them before?

submissions/logistics_belts/logistics-belts.md
```diff
@@ -0,0 +1,27 @@
...
...e "charging cloud" that comes about when your logistics network asks for no less than ten thousand gears and fifteen hundred electronic circuits. Likewise, when you ask for something like the materials to make a nuclear reactor or a rocket silo, bots will tend to swarm around an assembler in an area that was never designed for that degree of throughput. When you need a lot of an item, it's much easier to get it in by belt: sending ten thousand gears takes less than four minutes and barely damages your UPS; with bots it would take 2500 separate trips and likely make that many bots go airborne, swamping charging capacity that may never have been designed for that and taking forever to return to equilibrium. Running dedicated belts for every item, however, creates its own set of problems; often, the belts will be sitting idle and taking up space within your factory, and you may not even have room to run an eight-bus in the first place. For something like your belt mall, sending a large load of gears intermixed with a large load of electronic circuits is trivial for a logistics belt; once they arrive in that location, using bots internally is much less strain on your logistics network.
```
Conor12345, line 27: Could you add a larger scale example of the logistic belt concept in action here? You've done a great job explaining the basic but seeing where it can go would be super interesting.

Nanogamer7:
**Fact Checking:** 

Disclaimer: Since your piece is mostly about your own invention, my fact check will look rather lack luster, as I won't validate claims you've already proven

- ❔ You could argue inserter chains and moving items per hand are important methods to moving items too (don't take this too serious)
- ❔ [[Reddit](https://www.reddit.com/r/Allaizn/comments/90b3ac/optimizing_labs_for_ups/)(Source is mostly about belts, but entities are still listed and take more time to update] Unfortunately I couldn't find any newer sources directly comparing the UPS impact of bots vs. belts, but according to this post 2 years ago there are definitely cases where bots are better in terms of UPS than belts. [This](https://www.reddit.com/r/factorio/comments/k5y4rm/bots_vs_belts_and_ups_impact/) post from 2 months ago has some discussion in the comments, and while I don't necessarily want to draw any conclusions off of that, I'd still be careful with assumptions regarding bot UPS impact. That being said, it might be better to use belts for "ten thousand items"

---

✔ Correct
❗ Incorrect

Facts and Statement not listed above weren't validated by me; that doesn't mean they are false, it could be something personal (closed-source, etc.) I can't check. If you feel like a claim I left out should be validated, please tell me!

PR #395, opened 2021-01-31
Recipe-Oriented Factorio Life

pigheadx:
### Author info
Name: Ph.X

Still need more polish, but I would like to check some ideas from the editorial first. It would be great to have a better name, especially the **L** part.
The blueprints used for the pictures in this article are also attached. It requires LTN, AAI Containers & Warehouses, and Miniloader as mentioned in the article.
Discord: Ph.X#9753


Conor12345:
Firstly, the idea is awesome and is something I've personally wanted to try for a long time so I think there's plenty here to talk about. Here is feedback in hopefully chronological order:
 - Maybe explain how the red circuit one works a bit more specifically as the warehouse isn't very clear (probably just needs a caption on the image)
 - You have alot of examples of modules here which could maybe do with a bit more explaining
 - A more explicity TLDR of LTN is probably needed somewhere seen as it is a fundamental component of this. Doesn't need to long but needed to illustrate roughly how it works
 - I feel like it could maybe do with a conclusion on the end, possibly about using this method in future as a way to break down complex recipies, even if you arent using a massive rail network (probably a helpful idea for newer players)
 - We should (in theory) be able to make the map at the end into a zoomable google maps type thingy so if you could upload your save file somewhere and share it here we could have a go at that. and it might be a bit better than just a plain screenshot
 
I fear I may have some bias in reviewing this as its something I've thought about quite a lot in the past so some newer/less experience in this area reviews are probably needed to ensure it makes sense to everyone. Overall though, amazing piece!

pigheadx:
My screenshots may not be suitable for Alt-F4 in terms of resolution and zoom level. It would be great if there were better images to replace them. The blueprint is attached, please note that the orientation of the miniloader might be wrong.
The blueprints used in the save file can be found at https://github.com/pigheadx/fbb if needed.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
+## Recipe-Oriented Factorio Life <author>Ph.X</author>
+
+This paper proposes a universal factorio design philosophy. To better implement, [LTN - Logistic Train Network](https://mods.factorio.com/mods/Optera/LogisticTrainNetwork), [AAI Containers & Warehouses](https://mods.factorio.com/mod/aai-containers), and [Miniloader](https://mods.factorio.com/mod/miniloader) mods are being used.
+
+### What is ROFL
+
+People with programming experience should have heard of ~~[Object-Oriented Programming](https://en.wikipedia.org/wiki/Object-oriented_programming)~~ [Modular Programming](https://en.wikipedia.org/wiki/Modular_programming), and it is the theory which Recipe-Oriented Factorio Life (ROFL) mimics.
```
Conor12345: Maybe add a primer as to roughly what modular programming is here as it might help make the connection for some people as to the goal of ROFL. Maybe even include some kind of image to show how the flow of a modular programming project works - obviously would need to be a very simplistic version. This would also just be a nice opportunity to introduce people to modular programming who may have never heard of it before. Wooo learning!

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
+## Recipe-Oriented Factorio Life <author>Ph.X</author>
+
+This paper proposes a universal factorio design philosophy. To better implement, [LTN - Logistic Train Network](https://mods.factorio.com/mods/Optera/LogisticTrainNetwork), [AAI Containers & Warehouses](https://mods.factorio.com/mod/aai-containers), and [Miniloader](https://mods.factorio.com/mod/miniloader) mods are being used.
```
Conor12345: I'm not sure if this intro section here is necessary, feels a bit like a introduction for the sake of having one rather than just diving straight into the meat of the piece. Maybe move the mod list thing to the end of the what is ROFL section?

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+### What is ROFL
```
Conor12345: Hehehe good acronym :P

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+People with programming experience should have heard of ~~[Object-Oriented Programming](https://en.wikipedia.org/wiki/Object-oriented_programming)~~ [Modular Programming](https://en.wikipedia.org/wiki/Modular_programming), and it is the theory which Recipe-Oriented Factorio Life (ROFL) mimics.
+
+Like most philosophies, ROFL tries to find one eternal object as the base of everything, in this case, it is the recipe in Factorio. Red circuits can be crafted by an assembly machine of any tier, which may be affected by different modules, and the required raw materials may come from a belt, a logistics chest, or even an adjacent copper wire assembly machine, but no matter how other conditions may change, the recipe itself keeps being copper wires, electronic circuits with plastic bars to the advanced circuit.
```
Conor12345: Maybe mention that these are the smallest units of crafting and can't be divided any further. Not sure the exact wording of the concept I'm thinking of but something along the lines of the recipe can be subdivided no further.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+Based on a certain recipe, a dedicated subfactory is packaged into a module. Such a module contains everything necessary to process only one recipe of the desired factory. A separate logistics system is used to obtain raw materials and supply products for the modules. The purpose of this division is to achieve high cohesion inside the modules and low coupling between the modules.
```
Conor12345: > The purpose of this division is to achieve high cohesion inside the modules and low coupling between the modules.
- This feels like it says a lot in very few words which some people may not understand. I think its a good idea to expand this idea out and explain what you mean by high cohesion and low coupling. This seems like a very key concept of this piece so its a good idea not to rush it.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+In vanilla trains can only run on a predetermined schedule, which is not flexible and "logistical". This is where LTN comes in handy. In LTN, trains are parked in the depot stations when they are idle, just like logistic robots are placed in the roboports. Logistic train stop can acts like a requester chest or a provider chest depending on the signal they received. When there is a logistic demand, LTN will find a suitable train at the depots and generate a schedule for it to fetch the specified items at provider station, then go to the requester station for unloading, then return to a depot station, similar to the behavior of a logistic robot. But unlike logistic robots, the trains do not load or unload themselves, so inserters or loaders at the stations are needed. The LTN is powerful than the logistic robot network because of the huge cargo capacity, high speed, and fluid support.
```
Conor12345: - I think this is the time to talk about LTN specifically and link to the mod page etc.
 - This is a really good high-level description of how LTN works so good job there :D
 - Maybe mention what you mean by a depot - I have used the word before in a piece and had questions about it on the subreddit afterwards.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+The chosen rail grid in my practice is a two-way left-hand traffic railroad in order to place the rail signals between the rails. Junctions are straight and left turn only for space-saving. Most stations have a train stop limit of 2 to ensure that a maximum of one train waits outside the station. Sections occupied by a handful of trains waiting outside the stations can be detoured through the fully interconnected grid.
```
Conor12345: Could you include a blueprint string/book for the rail and modules you are using for the rest of the piece in case anyone wants to play with it?

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+![rail grid](figure/rail_grid.jpg)
+_Only the center grid will build a module, the rest grids are used for traffic and solar paving_
```
Conor12345: Think this caption needs some minor wording tweaks to get across what youre trying to say
 - will ~~build~~ contain a crafting module
 - used for train traffic management and solar power generation

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+![Gear module](figure/rofl_gear_module.jpg)
+_Simply gear module. Raw material station is below, product station is above_
+
+Since the inputs and outputs within this module are fulfilled by trains, our modification and upgrade of the production line within the module will not affect other modules.
+
+![Alt gear module](figure/rofl_gear_module_2.jpg)
+_Gear module with beacons. No change of stations, only the production line between stations._
+
+Next, let‘s look at an advanced circuit module with three types of raw materials. By setting the logistic train stop requesting multiple ingredients, we were able to unload all of our raw materials at a single station. At this point all the raw materials are mixed in one warehouse， we then use filter miniloaders (similar to filter inserters) to separate each material onto a separate belt to supply the assembly machine.
```
Conor12345: Maybe expand on what warehouses and loaders are for anyone not familiar with the concepts. This could also be the point to work in the link to the loader and warehouse mods. Also probably worth saying that is can be done without these mods - possibly could use a compare image thing to show a version using the mod and one not using the mods - see https://alt-f4.blog/ALTF4-X/#compare-images

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+![Oil refinery module](figure/oil_refinery_module.jpg)
+_Oil refinery module. Note that by controlling the pump via the LTN signal allows different fluids to be loaded or unloaded at the same station without mixing._
+
+There are also some unconventional recipes, such as research, thermal power plant, and rocket silo.
+
+![research center module](figure/research_center_module.jpg)
+_Research center module, which doesn't need an output station_
+
+![thermal power plant module](figure/thermal_power_plant_module.jpg)
+_Thermal power plant with landfill to place on water._
```
Conor12345: Maybe a good point to mention a mod like https://mods.factorio.com/mod/LandfillEverything. Not essential but its always good to highlight cool mods when we can.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+![rocket silo module](figure/rocket_silo_module.jpg)
+_Rocket silo module. Independent logistics robot network can be used within the module. The bottom station is for materials needed to build rockets, and the station on right is for payloads of the rocket, usually satellites._
```
Conor12345: Oooo this is very cool having just completed a high SPM base myself where I utilized separate logistics networks a lot. Maybe mention why someone would want to do this - the advantage of high throughput of items without running a million belts and loading balancing etc. Careful not to get too side tracked though since its not that key to the understanding of the piece.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Pros and cons of ROFL
+
+ROFL is optimized towards engineers, making it easy for players to divide and conquer production problems. When you need to implement a certain recipe, simply set up input and output stations in the grid that match the types of raw materials and products, and then design production lines between the stations. When the capacity of a recipe is insufficient, you can directly copy and paste the corresponding factory module and integrate it into the whole factory production without [plumbing up the inputs and outputs](https://alt-f4.blog/ALTF4-22/#the-goal-of-factorio).
```
Conor12345: :D thanks for the sneaky plug here

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
```diff
@@ -0,0 +1,75 @@
...
+
+ROFL is not optimized towards production, it is not a perfect-ratio production line, so it's common that structures stop working because of insufficient raw materials or product overflow. Each item has to be loaded to and unloaded from a wagon, adding a lot of input and output bottlenecks, most typically in copper wire, where the number of products exceeds the number of raw materials and the number of stacks is high while the demand is huge. There is a very significant cache with the warehouses, which is not conducive to identify the items with insufficient production in time (definitely iron ore though). The rail network used for inter-module logistics in the factory is expensive to build and also poses a large number of traffic accident hazards (praise the spidertron).
+
```
Conor12345, line 62: Maybe add a section here about how you can use the concepts of ROFL in a less extreme way to divide and conquer complex recipes into more manageable chunks. This could make the piece more directly useful for newer players that may get overwhelmed by the complexity of some end game recipes.

Conor12345 (commented):
I really like your piece having just tried to do this myself in my most recent base, though with much more spaghetti. I've left quite a lot of specific feedback but don't feel like you need to implement all the changes I suggested  if you don't agree with them, it is your piece after all :D.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Added a description of modular programming from Wikipedia and the ROFL equivalent.
I agree that a picture would describe the concept better, but I couldn't find or draw a picture that would be understandable to a non-programmer.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Removed the paragraph for now. I may try to write a better introduction later.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Proud of it, but as I wrote in my initial comment, I‘m looking for a more appropriate L word.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Feels like "atomic operation", but the term is used more in the realm of synchronization or multi-threading related areas and may not fit here.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Added a specific explanation of high cohesion and low coupling.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Move the mod declaration and link to the end of What is ROFL.
"trains parked when they are idle " should explain the depot I think, Nevertheless, the sentence has been slightly altered.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Blueprint link added to the end of What is ROFL.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Edited as suggested

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Further explanation of such popular mods as Warehouse and Loader may be a bit redundant to me, especially since such mods have appeared in Alt-F4 in the past. 

Adding a vanilla compare to a modded loading station is a fascinating idea, it's just that I'm currently having trouble making relevant media content and I need help to make pictures.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: Can't believe that I missed such a great mod. Definitely install it in the next playthrough.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: The advantages of separate logistics networks are described after the rail grid, around line 26.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: In fact, this is one of the reasons that motivated me to write this article, since I want to prove that blueprints can be even more hassle-free.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: I wrote about extending ROFL in the Conclusion section, but it is quite abstract. I think it doesn't need more generalized examples, focus on explaining the modularization idea is enough.

submissions/recipe_oriented_factorio_life/recipe_oriented_factorio_life.md
pigheadx: After rethinking, I realized that it was impossible to simply and elegantly reproduce a station with similar functionality in vanilla. The main difficulty is that LTN can control the quantity of different raw material items and be able to unload multiple types of items in one station without stoppage. Vanilla, on the other hand, cannot distinguish different items even with the 1.1 train stop limit, and there is a high risk of an overflow of one ingredient and jams the traffic.

Nanogamer7:
**Fact Checking:** 

As you aren't presenting an existing, but a new concept, there isn't much to fact check, however I still listed the few minor details I found below

- ❗ *"In vanilla, trains can only run on a predetermined schedule, which is not flexible and 'logistical'."*: With train stops having the same name and some circuit magic, I'd argue that there's definitely some flexibility and logistically in unmodded factorio, which goes as far as allowing pull-based logistics systems just with vanilla, as Conor already explained in detail in [issue 15](https://alt-f4.blog/ALTF4-15/#train-stop-limits-conor_). Although I have to agree, that LTN allows for even more flexibility, or at the very least makes it more intuitve.

---

✔ Correct
❗ Incorrect

pigheadx:
@Nanogamer7 
I've thought about implementing similar functionality in vanilla. For a single item type, it is easy to implement LTN-like effects by controlling train stop limits with signals and using double depot schedules. However, the problem arises when using a single station to unload multiple item types. Train stop limits can only control the number of trains but not the type of train, which usually results in one item being piled up in the warehouse while the others cannot be unloaded. If we want to control the balanced unloading of multiple items in vanilla, we may need a global signaling network and control train stop limits on both the supply and demand side, which is basically equivalent to re-implementing LTN with an in-game signaling network, far beyond my capabilities. For my design, the vanilla train is really not flexible enough.

Nanogamer7:
Fair enough. I'd still prefer to rephrase this sentence, not only in the context of ROFL, but the statement itself; if you have any suggestions, feel free to do it yourself, otherwise we can handle it on the editorial side ;)

pigheadx:
I'm sure you can describe this fact better, so I'll leave it to the editorial team.

PR #396, opened 2021-02-08
Belt Computers Part 1

Conor12345:
### Author info
Name: pocarski
Suggested release date: 12/02/2021


Nanogamer7:
**Fact Checking:** 

- ❗ Your link to DOOM in Factorio directs to the first video of the playlist. Since it isn't an introduction to the project, I'd prefer a link to the playlist: https://www.youtube.com/playlist?list=PLdkxm81SV4uVs7EHAxeGcaPLe9xX_7tjb
- *Validation for your description of a turing machine will follow soon*
- ❔ I don't think the blueprint book string should be hosted in this way on an external site, however as we don't have a better solution for now, it will do
- ✔ [[Wikipedia](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)] You only need a NOT gate and either an AND or an OR gate to do all binary logic
- ✔ [[Wikipedia](https://en.wikipedia.org/wiki/Logic_gate)] All logic gates correspond to the definitions on the wiki-page
- ✔ [[Wikipedia](https://en.wikipedia.org/wiki/Adder_(electronics))] The half-adder and full-adder correspond to the definitions on the wiki-page
- ✔ The compacted adder is *"exactly the same circuit as in the previous image"*, and the fact checker verified that ;)
- ✔ All other links direct to where they say they are going.

---

Further notes:
- I'd love to see videos of the logic gates in action (changing the inputs and showing the resulting changes in the outputs), though I understand if this is to troublesome for now

---

✔ Correct
❗ Incorrect

Facts and Statement not listed above weren't validated by me; that doesn't mean they are false, it could be something personal (closed-source, etc.) I can't check. If you feel like a claim I left out should be validated, please tell me!

PR #402, opened 2021-02-11
remove modernizr-webp.js

justchen1369:


justchen1369:
hmm, it appears that FA is used in the buttons on the bottom of the site.

DrMagPie:
Language button 
arrows in page selector

Conor12345:
Would this be a good time to switch to self hosting the primary fonts? I remember there was something about it being more beneficial since browser privacy rules have chanced. Also probably worth finally removing google analytics which I completely forgot to do.

Conor12345:
I believe this is the font sled hosting thing I was thinking of: https://wpspeedmatters.com/self-host-google-fonts/

TLDR a primary advantage of Google Fonts was they were likely pre cached from other sites but this can't happen anymore cause cache segmenting. Google Fonts could be faster but it also adds more DNS checks and stuff so I suspect hosting the fonts ourselves would be faster

justchen1369:
hmm, yeah. @YPetremann comments?

YPetremann:
I don't see any good argument and counter-argument for the need to move to self host font-awesome.

CanadaHonk:
Imo:
Self-hosting font awesome isn't strictly needed however self-hosting core fonts for the page is a good idea.

Zr4g0n:
The icons that are used from FA are simple enough that they could be replaced by inline SVGs instead.

justchen1369:
based on ctrl+f, there are 11 FA icons that are used, so I don't think we should remove it.

PR #412, opened 2021-02-19
Small SEO improvements

CanadaHonk:
- Removes duplicate title HTML tag
 - Removes duplicate generator meta tag


justchen1369:
This seems to do no harm, though lighthouse reports no difference in SEO.

PR #429, opened 2021-03-16
Complexity Corner 4

jmkool:
### Author:


submissions/Complexity Corner 4/Complexity Corner 4.md
```diff
@@ -0,0 +1,27 @@
...
+
+What are beacons?
+
+The [Beacon](https://wiki.factorio.com/Beacon) is a building that takes up a space of 3x3 tiles.  It first appears relatively late in your usual playthrough, requiring purple science to research, but the most advanced ingredient is red circuits.  A beacon has two module slots, and will transmit half the effect of those modules across a 9x9 area, reaching anything within 3 tiles of the beacon.  All this is to say, it takes a module and spreads its effect across nearby buildings.
+
+What results can they give?
+
+Most people see beacons as a megabase tool, and it's true, they are very effective at improving UPS (Updates Per Second).  With faster machines, you need fewer of them, and fewer things moving means your computer can handle bigger factories.  Improving UPS is far from they only thing they do, though.
```
heinwessels: > Improving UPS is far from they only thing they do, though, and can have uses in smaller bases as well.

Could also mention somewhere to watch out for their big power draw.

submissions/Complexity Corner 4/Complexity Corner 4.md
```diff
@@ -0,0 +1,27 @@
...
+
+A quick note must be given to what beacons cannot do: Beacons cannot contain productivity modules, and beacons will only boost a building that has module slots.  So no, you can't put speed modules in a beacon and expect your steel furnace to work faster.  Your electric furnace, however, would work faster.
+
+Where should they be used?
+
+It usually doesn't make sense to put efficiency modules in beacons, but usually doesn't mean always.  If the energy reduction on affected buildings would offset the energy usage of your beacon, then you get a net energy savings.  While this is unlikely to happen with most buildings, each beacon needs only reach three Assembly Machine 3s to achieve this if you're using Efficiency 3 modules, so it's worth keeping in mind.
```
heinwessels: How does the power saved using efficiency modules compare to the staggering 480MW that a beacon draws? Does this mean it's only useful on certain machines?

submissions/Complexity Corner 4/Complexity Corner 4.md
```diff
@@ -0,0 +1,27 @@
+### Complexity Corner: Beacons <author>The Kool</author>
+
+Welcome back to Complexity Corner!  This week, I'll be taking a slight left on the topic of modules that I've held for the first three articles, and will be going over the beacon, a building completely reliant on modules for its existence.  As before, I will ask three questions about this potentially unfamiliar thing: “What is it?”, “What results can it give?”, and “Where should it be used?”  My goal is to show those of you who haven't used it yet just how easy it is to work with.
```
heinwessels: `detour` might be a better word than `left`

submissions/Complexity Corner 4/Complexity Corner 4.md
```diff
@@ -0,0 +1,27 @@
...
+
+What are beacons?
+
+The [Beacon](https://wiki.factorio.com/Beacon) is a building that takes up a space of 3x3 tiles.  It first appears relatively late in your usual playthrough, requiring purple science to research, but the most advanced ingredient is red circuits.  A beacon has two module slots, and will transmit half the effect of those modules across a 9x9 area, reaching anything within 3 tiles of the beacon.  All this is to say, it takes a module and spreads its effect across nearby buildings.
```
heinwessels: The first sentence should read something like: beacons are machines that effects other machines in close range. Then you can start explaining more technical things like the size, etc.

heinwessels (commented):
Good work, as always! Not much to add this time. We would likely want this piece for next week. :)

submissions/Complexity Corner 4/Complexity Corner 4.md
jmkool: I should improve this to show some numbers a bit more clearly.  That 'staggering' 480MW may sound like you can never recover it, but I actually did crunch the numbers.  If you place two Eff3 modules in a beacon, and that beacon reaches buildings with a combined base draw of 960+MW, then you are actually reducing power usage by placing the beacon.  For example, if the beacon reaches at least three Assembly mk3 machines.

heinwessels: That's a great point to add.

PR #453, opened 2021-04-12
Pocarski - Combinators 

heinwessels:
pocarski


submissions/pocarski-combinators/pocarski-combinators.md
```diff
@@ -0,0 +1,47 @@
...
+
+(counter)
+
+Finally, the counter. Structurally it is identical to the SR latch, but this time the output is set to "input count of everything". This means that, while the decider's condition is followed, it will keep giving its own outputs to itself, thus remembering them. For every tick it receives a signal, it will increment the value of that signal in its memory by the amount received. As soon as the condition is broken, the memory is cleared, since the decider no longer allows signals to pass. Similarly to the counter, if you wish to make it remember only one signal, replace the "everything" in the output with the desired signal. This circuit, just like the monostable, has an immense number of uses, but the most popular one is to keep track of item amounts.
```
DaleStan: I think this should say "identical to the RS latch", not SR latch.

submissions/pocarski-combinators/pocarski-combinators.md
```diff
@@ -0,0 +1,47 @@
...
+
+(backup switch)
+
+Finally, a process that many fear to set up: uranium enrichment. The part we're worried about is the recycling of uranium. We want to stop recycling U-235 as soon as we hit 40 items, after which the extra one should be placed into the output. Naturally, we use a counter for this. The recycling inserter is set to "read hand - pulse", so every time it grabs uranium for recycling the counter counts up. As soon as the amount in the counter becomes greater or equal to 40, the recycling inserter's filter is changed to U-238. U-235 has higher signal priority, so this can be done by simply turning off "U-235" while there is a constant input of "U-238". This means it takes out U-238 on its fifth swing, which resets the counter and triggers the U-235 output inserter. We don't need any special handling of the recycled U-238 because the input inserter's stack size is 3, meaning there will never be any competition between U-238 inserters, since their respective item counts add up to 5, the exact amount needed.
```
DaleStan: It may be better when accompanied by an annotated image, but this last sentence seems too long.

Nanogamer7:
**Fact Checking:** 

- ❗ The term "monostable" itself doesn't necessarily refer to a pulse generator (though it can). What you were descriping fits the definition of a Monostable_multivibrator [[Wikipedia]](https://en.wikipedia.org/wiki/Monostable_multivibrator) (not an expert though, don't take my word for it), but for example something that defaults to, lets say, 7 iron, would be a monostable circuit aswell, 7 iron being the stable position. (A practical usecase for such circuits would be requests, the desired amount being the stable postion, and the current storage changing this to whatever is needed). Although it has to be said there isn't a fixed definition for monostable (especially in context of Factorio), so there's definitely room to argue.
- ❗ See DaleStans comment, both (SR- and RS-latch) appear to be valid terms, but please be consistent within your article (image place holder, latch itself, and 'reactor fueling' use "rs"; 'counter' uses "sr")
- ✔ [Wikipedia](https://en.wikipedia.org/wiki/Flip-flop_%28electronics%29#SR_NAND_latch) Your explanation fits the wikipedia of flip flops/latches
  - ❔ An explanation to the letters "S" and "R" as set and reset would make it easier to follow, especially for the backup switch
- {more will follow with images}

---

✔ Correct
❗ Incorrect

PR #471, opened 2021-04-28
The Engineer's Physics-Defying Strength

heinwessels:
author: Romner


submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
+### The Engineer's Physics-Defying Strength <author>Romner</author>
+
+Minecraft's Steve is strong.  
+Lift-billions-of-kilograms-with-no-problems-strong.
```
heinwessels: ~Why? Is there something we can link to show why Steve is so strong?~
Oh, I see you gave a link at the end. I do think it would be put here though, otherwise, it seems like you're pulling numbers out of thin air.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
+### The Engineer's Physics-Defying Strength <author>Romner</author>
+
+Minecraft's Steve is strong.  
+Lift-billions-of-kilograms-with-no-problems-strong.  
+But he doesn't even *compare* to Factorio's Engineer.
+
+Taking the maximum of 120 inventory slots + 30 trash slots in vanilla and putting 10 nuclear reactors in each gives us 1500 reactors.
```
heinwessels: This is said a little hastily without describing what you're doing and it's very easy to miss the goal. You need to mention you're going to calculate how strong the engineer is by approximating how much weight he can carry using Nuclear Reactors. Something like:

> How much can the engineer carry? Let's assume the engineer is wearing Power Armor Mk2, which gives a total of 120 inventory slots. Then he also has 30 Logistic Trash slots which can be filled with items. We can approximate how much he can carry by filling all these slots with the heaviest item in the game: the Nuclear Reactor, which stacks to 10.  So the engineer can carry a total of 1500 reactors at once. According...

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+$9\ 593\ 600 + 109\ 367\ 040 = 118\ 960\ 640$ metric **tons,** or **118 960 640 000 kilograms.**  
+
+Now, most people would be satisfied with that number; it is much higher than what Steve can carry.  
+But we can do better.  
+Much better.
+
+Remember [ALT-F4 #11](https://alt-f4.blog/ALTF4-11/), specifically the "[Nauvis is a white dwarf](https://alt-f4.blog/ALTF4-11/#nauvis-is-a-white-dwarf)" article by pocarski?
+
+In that article, he calculated that the density of Nauvis has to be at least 813 400kg/m³ for itself to not fall apart.  
+Earth has 5 510kg/m³, around 147.6225x less; which, assuming that Nauvis is comparable to Earth in terms of size, also means around 147.6225x the gravity. Which means that Factorio's Engineer can carry:
```
heinwessels: When making comparisons keep it to easy to read numbers:

> around 150 times less

Instead of

> around 147.6225x

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
+### The Engineer's Physics-Defying Strength <author>Romner</author>
+
+Minecraft's Steve is strong.  
+Lift-billions-of-kilograms-with-no-problems-strong.  
+But he doesn't even *compare* to Factorio's Engineer.
+
+Taking the maximum of 120 inventory slots + 30 trash slots in vanilla and putting 10 nuclear reactors in each gives us 1500 reactors.  
+According to [this article,](https://www.nextbigfuture.com/2007/07/constructing-lot-of-nuclear-power.html) modern nuclear reactors use 40 metric tons of steel and 190m³ of concrete per MW.
+
+Nuclear reactors in Factorio produce 40MW. Each. Oh, and there's also the neighborhood bonus, which we can account for with an equation (assuming 4 or more reactors and a 2xN reactor layout), where $n$ is the number of reactors and $P$ is the base power output of the reactor:
```
heinwessels: > Nuclear reactors in Factorio produce 40MW each. There is also the ~neighborhood~ neighbour bonus, which can be accounted for with...

heinwessels: This should be part of the previous paragrapgh.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+Lift-billions-of-kilograms-with-no-problems-strong.  
+But he doesn't even *compare* to Factorio's Engineer.
+
+Taking the maximum of 120 inventory slots + 30 trash slots in vanilla and putting 10 nuclear reactors in each gives us 1500 reactors.  
+According to [this article,](https://www.nextbigfuture.com/2007/07/constructing-lot-of-nuclear-power.html) modern nuclear reactors use 40 metric tons of steel and 190m³ of concrete per MW.
+
+Nuclear reactors in Factorio produce 40MW. Each. Oh, and there's also the neighborhood bonus, which we can account for with an equation (assuming 4 or more reactors and a 2xN reactor layout), where $n$ is the number of reactors and $P$ is the base power output of the reactor:  
+$(n - 4)*P*4 + 4*P*3$  
+Using this, we get a total of...  
+$(1500 - 4)*40*4 + 4*40*3 = 239\ 840\mathrm{MW}$
```
heinwessels: `239.84 GW`

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+Nuclear reactors in Factorio produce 40MW. Each. Oh, and there's also the neighborhood bonus, which we can account for with an equation (assuming 4 or more reactors and a 2xN reactor layout), where $n$ is the number of reactors and $P$ is the base power output of the reactor:  
+$(n - 4)*P*4 + 4*P*3$  
+Using this, we get a total of...  
+$(1500 - 4)*40*4 + 4*40*3 = 239\ 840\mathrm{MW}$  
+With the numbers from the [earlier article](https://www.nextbigfuture.com/2007/07/constructing-lot-of-nuclear-power.html) and a [quick google search](https://www.google.com/search?q=concrete+density+kg%2Fm3&oq=concrete+density&aqs=edge.1.69i57j0l2j0i20i263j0l3.13097j0j4&sourceid=chrome&ie=UTF-8), we can calculate the total mass:  
+$239\ 840 * 40 = 9\ 593\ 600$ metric tons of steel and  
+$239\ 840 * 190 * 2.4 = 109\ 367\ 040$ metric tons of concrete, equals a total of  
+$9\ 593\ 600 + 109\ 367\ 040 = 118\ 960\ 640$ metric **tons,** or **118 960 640 000 kilograms.**  
+
+Now, most people would be satisfied with that number; it is much higher than what Steve can carry.
```
heinwessels: I think you need to mention what it is you are doing next, otherwise, it's very easy to lose the reader. Maybe something like:

> However, we have only calculated all the nuclear reactor's mass and not the weight! This is already very heavy for an item on earth, but what about the gravity of Nauvis? Remember Alt-F4 #11...

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
heinwessels: Expand a little on your sentences here. Something like:

> On the other hand, Earth has a density of 5510kg/m3, which is around 150 times less, assuming... Therefore, this means Nauvis also has 150 times higher gravity than the Earth.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+But we aren't done yet, nonono.  
+You see, using pocarski's equation on Earth gives us around:  
+$\frac{3\pi}{G\ *\ 86\ 400^2} = 18.9\mathrm{kg/m^3}$, which is ~290x less than Earth's actual density of 5 510kg/m³.  
+Trying it on Mars gives us a similar result:  
+$\frac{3\pi}{G\ *\ 88\ 619.616^2} = 17.98\mathrm{kg/m^3}$, which is ~220x less that Mars' density of 3 930kg/m³.  
+This means that an Earth-like planet should have around 200-300 times the density calculated by pocarski's equation.
+
+That means Nauvis' density is closer to $813\ 400 * 250 = 203\ 350\ 000\mathrm{kg/m^3.}$  
+Calculating the total weight with that value gives us:  
+$203\ 350\ 000 : 5\ 510 * 118\ 960\ 640\ 000 = 4\ 390\ 316\ 904\ 537\ 205\mathrm{kg.}$
```
heinwessels: Same thing again.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
heinwessels: When you're multiplying numbers, rather use `x`, it's much easier to read. You can do this using `\times`.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+
+Now, most people would be satisfied with that number; it is much higher than what Steve can carry.  
+But we can do better.  
+Much better.
+
+Remember [ALT-F4 #11](https://alt-f4.blog/ALTF4-11/), specifically the "[Nauvis is a white dwarf](https://alt-f4.blog/ALTF4-11/#nauvis-is-a-white-dwarf)" article by pocarski?
+
+In that article, he calculated that the density of Nauvis has to be at least 813 400kg/m³ for itself to not fall apart.  
+Earth has 5 510kg/m³, around 147.6225x less; which, assuming that Nauvis is comparable to Earth in terms of size, also means around 147.6225x the gravity. Which means that Factorio's Engineer can carry:  
+$813\ 400 : 5\ 510 * 118\ 960\ 640\ 000 = 17\ 561\ 267\ 618\ 148\mathrm{kg}$ or so worth of nuclear reactors.
```
heinwessels: I'm not sure I understand this equation. What does the `:` do? I assume you meant a divide `/`. In that case, you're redoing math you already did. Simply say `147.6625 x 118..... = 17....`

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+
+**17 561 267 618 148 kilograms.  
+Seventeen and a half *trillion* kilograms.**
+
+But we aren't done yet, nonono.
```
heinwessels: I'm not sure about this line. Do you think you can try to rewrite it a little? You can mention that the previous calculation was based on the assumption that pocarski's density equation is accurate. However, this equation is very conservative by design, and the actual density is most likely much more!

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+
+In that article, he calculated that the density of Nauvis has to be at least 813 400kg/m³ for itself to not fall apart.  
+Earth has 5 510kg/m³, around 147.6225x less; which, assuming that Nauvis is comparable to Earth in terms of size, also means around 147.6225x the gravity. Which means that Factorio's Engineer can carry:  
+$813\ 400 : 5\ 510 * 118\ 960\ 640\ 000 = 17\ 561\ 267\ 618\ 148\mathrm{kg}$ or so worth of nuclear reactors.
+
+**17 561 267 618 148 kilograms.  
+Seventeen and a half *trillion* kilograms.**
+
+But we aren't done yet, nonono.  
+You see, using pocarski's equation on Earth gives us around:
```
heinwessels: > using pocarski's equation for density on Earth gives us

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+In that article, he calculated that the density of Nauvis has to be at least 813 400kg/m³ for itself to not fall apart.  
+Earth has 5 510kg/m³, around 147.6225x less; which, assuming that Nauvis is comparable to Earth in terms of size, also means around 147.6225x the gravity. Which means that Factorio's Engineer can carry:  
+$813\ 400 : 5\ 510 * 118\ 960\ 640\ 000 = 17\ 561\ 267\ 618\ 148\mathrm{kg}$ or so worth of nuclear reactors.
+
+**17 561 267 618 148 kilograms.  
+Seventeen and a half *trillion* kilograms.**
+
+But we aren't done yet, nonono.  
+You see, using pocarski's equation on Earth gives us around:  
+$\frac{3\pi}{G\ *\ 86\ 400^2} = 18.9\mathrm{kg/m^3}$, which is ~290x less than Earth's actual density of 5 510kg/m³.
```
heinwessels: Rather write out `around` instead of using `~`

heinwessels: Why is it different? I think you should add at least one sentence as to the reason. I think it's because pocarski's equation assumes a uniform composition which Earth does not have.

submissions/engineers-physics/The_Engineers_Physics-Defying_Strength.md
```diff
@@ -0,0 +1,51 @@
...
+Calculating the total weight with that value gives us:  
+$203\ 350\ 000 : 5\ 510 * 118\ 960\ 640\ 000 = 4\ 390\ 316\ 904\ 537\ 205\mathrm{kg.}$
+
+**4 390 316 904 537 205 kilograms.  
+Almost *four and a half quadrillion kilograms.***
+
+For comparison, that's about the total weight of *[everything ever made by humans](https://www.nationalgeographic.com/environment/article/human-made-materials-now-equal-weight-of-all-life-on-earth), **four times.***  
+Or, the weight of [all living things on the Earth](https://en.wikipedia.org/wiki/Biomass_(ecology)#:~:text=The%20total%20live%20biomass%20on%20Earth%20is%20about%20550%E2%80%93560%20billion%20tonnes), *eight times.*
+
+Steve can carry around [5 billion kilograms.](https://www.reddit.com/r/Minecraft/comments/jmz6lz/actualized_maximum_weight_that_steve_can_hold/) That is around *870 000 times less*.
```
heinwessels: It sounds like you are bashing Minecraft a little. Maybe rather simply focus on how strong the engineer is, and not that Steve is weak, because he definitely is not. You can say something like `at least Steve can break trees and rocks with his bare fists`. Or `maybe that is why Steve can jump and even climb massive mountains, while the engineer cannot even get past a little cliff.` Most of us are Minecraft fans too :P

heinwessels (commented):
Hello Romner! Really nice article! I always like something with a bit of physics that pulls things into weird extremes. It looks like it has a lot of comments, but most of it is simply cosmetic. I think this will be another great sciency article for Alt-F4! Thanks!

Nanogamer7:
**Fact Checking:** 

- ~~For the purposes of this article I am using the word *"strong"* as in *"can carry in survival / without cheats"*, although there's an argument to be made to consider total strenght (see the [discussion](https://discord.com/channels/745990677606826005/748304365831323830/835892828860317716) on discord for more)~~ Point adressed at the end of the article
- ✔ [[WIKI](https://wiki.factorio.com/Player)] The maximal carrying capacity of the enineer is 120+30 slots
  - It should be noted, however, that this requires the Toolbelt research. For the purposes of this article we can assume though, that the average reader is aware of this and/or doesn't care
- ✔ Your numbers are accurate to the linked article
- ❔ Your calculations appear to make sense, but unfortunately I can't reconstruct if you are using them in the right context. I'd ask you to check them again later, if you notice any problems you missed previously
- ✔ 20 times higher = much higher
- ✔ [Newton's law of universal gravitation] Increasing mass increases by the same amount
- ❗ The *'actual density'* part could use a *"speculation"* disclaimer, as this is just based in the few similar planets we know about. (Don't take this point seriously)
- ❗ Line 37 uses an asterix `*` instead of the mathjax symbols
- ✔ Compared numbers are accurate to their linked articles

---

✔ Correct
❗ Incorrect

Conor12345 (approved):

PR #475, opened 2021-04-29
added gallery

### LINKED ISSUE #461: Gallery photo embed
I think it would be cool to have an embed which contains a group of images which the user can scroll through. It would be ideal for showing lots of images without making massively long pages.

Probably something where there is side to side buttons below the image to switch between images. Also probably needs a thumbnail view type thing to see all the image previews?

Hopefully there's a jekyll plugin that can do it relatively plug and play

DrMagPie:
Closes #461

Not tested on mobile

Posible improvements:
- Make multiple Galleries independent of each other.
- CSS animations for smoother transitions.


Conor12345:
This looks really nice! I think the basic bones are there it just could do with a couple little tweaks:
 - `galery` -> `gallery` 🤣 
 - Could you have all the images show in the row, including the one already selected?
 - Can the row be moved to below the image? Also maybe centered?
 - I feel like it needs a bit of text along the lines of "click thumbnails to view other images"
 - Thumbnails may need to be a bit bigger
 - Add some padding around each image, especially between the large image and the smaller ones
 
 Feel free to disagree with any of the above. This looks like a really great feature :D
 
 *Note to future Conor : I think mobile is gonna be fine based of dev tools but it's probably worth doing a proper mobile test before merging*

DrMagPie:
Did some updates.

Regarding points above:

> -  galery -> gallery 🤣

it was totally intentional.... 🤣 

> - Could you have all the images show in the row, including the one already selected?

it is optimisation question. Currently it is made in a way that all images are loaded once and selected one just moves to bigger container in DOM

> - Can the row be moved to below the image? Also maybe centered?

Below: done.
Centred: well.... due to some css grid stuff I'm not sure how to do it. But is should be possible with flexbox, if some one wants to restyle it.

> - I feel like it needs a bit of text along the lines of "click thumbnails to view other images"

I did think about it. But at the moment I don't see how to put it there organically(so it would look in place).
EDIT:
Might be some kind of popup/tooltip with css. 
Although might be not... just tried it and no luck :(

> - Thumbnails may need to be a bit bigger

Increased min size to 100px. 

> - Add some padding around each image, especially between the large image and the smaller ones

Added gaps.

Conor12345 (approved):
Looks good - working fine on mobile too 👍

Conor12345:
Need to add caption support to this embed too

heinwessels:
This looks good to me too. Thanks DrMagPie! I think we can merge this.

PR #493, opened 2021-05-16
Audio Embed

Conor12345:
Please test and lemme know if you can break it.

### DIFF
--- a/_includes/audio.html
+++ b/_includes/audio.html
@@ -1,0 +1,7 @@
+<p></p> <!-- Spacing -->
+<div class="audio">
+    <audio controls>
+        <source src="{{ include.src }}" type="audio/mpeg">
+    </audio>
+</div>
+<p></p> <!-- Spacing -->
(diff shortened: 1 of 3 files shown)


heinwessels:
Is it possible to change the white to gray? The white is a bit drastic. I was thinking it would look nice if it was a darker gray than the background. But other than that it looks awesome :D

DrMagPie:
yeah I will agree with what @heinwessels said.

it would be grate to have it look consistently across browsers

firefox 
image

safari
image


besides that, from functional point it looks great

Conor12345:
So it looks like if we want to change the styling to be consistent, we have to start fresh and redefine the styling of the `<audio>` tag. (See https://stackoverflow.com/questions/4126708/is-it-possible-to-style-html5-audio-tag/4126871). This is probably beyond my js ability but if someone else can take a crack at it that would be awesome.

Another option would be to use something like http://www.jplayer.org/ which might make it a bit easier.

heinwessels:
Not sure if we still need it. Can we close it? We can always reopen it again when we want.

DrMagPie:
I believe that's for @Conor12345 to decide. As he is a "story" owner.

heinwessels:
Closing this PR as it clutters my `Review Reqeusts` page, and I'm pretty sure we're not going to merge it. We can always reopen it again if Conor wants to merge :)

PR #529, opened 2021-07-20
moved footer lower

### LINKED ISSUE #528: Spidertron blocking buttons on mobile
On mobile the spidertorn blocks the lower page buttons from being clicked on some screens. If you move the spider away it's still connected, so instead of clicking on the button the spider just moves to where you click. And when you deselect the spider to click the button while he is away he runs back to its home and over the button again. 

Possible workarounds:
- When deselecting spidertron leave it at last location.
- Spidertron is selected by double-tap. When single tapping it should still check and trigger the button behind the spidertron.
- I think we just need to add a bit more space between the body and footer (only a temporary measure?)

DrMagPie:
closes #528

### DIFF
--- a/_sass/footer.scss
+++ b/_sass/footer.scss
@@ -94,8 +94,10 @@
 }
 
 @media (max-width: 543px) {
-  .footer .footer-inner {
-    display: block;
+  .footer {
+    margin-top: 80px;
+    .footer-inner {
+      display: block;
   }
 
   .footer-inner .footer-links {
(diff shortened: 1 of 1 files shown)


xthexder:
Seems good, though it can still block the page selection buttons on the homepage at some screen sizes.
I might try and make the active area for the spidertron a little smaller, then maybe we can just add a small space for all screen sizes.

Conor12345 (approved):

PR #530, opened 2021-07-20
Update spidertron to minimize padding

xthexder:
Related to https://github.com/AlternativeFFFF/Alt-F4/pull/529, this makes the clickable area around spidertron much tighter, which should help with resolving https://github.com/AlternativeFFFF/Alt-F4/issues/528

screenshot


DrMagPie:
this is much better.
although, in some cases still abstracts most of the button. 
image

DrMagPie (approved):

PR #532, opened 2021-07-22
Issue 43 - Chinese Translation

pigheadx:
Issue 43 - Chinese Translation


Firerazer (approved):
ok if transcription not translated on purpose

PR #523, opened 2021-07-09
Trainperspective

arrow-in-my-gluteus-maximus:
Name: arrow in my gluteus maximus

PR #535, opened 2021-07-23
Juxtapose

DrMagPie:
Juxtapose requested in #523


arrow-in-my-gluteus-maximus:
:-O

heinwessels (changes requested):
I'm not 100% sure what the difference is, but it seems that #2 is centered better vertically, so I'll say that one. Will the caption be in the middle then? I think one caption for the entire thing is fine.

But it looks good, thanks!


Could you add the example to the `AltF4-X` example page.
https://github.com/AlternativeFFFF/Alt-F4/blob/master/_drafts/2020-08-20-ALTF4-X.md

heinwessels (approved):

PR #543, opened 2021-08-08
added default thumbnail thingy

### LINKED ISSUE #540: Superfluous asset folders
It's a bit annoying that the thumbnails each need their own folder in [here](https://github.com/AlternativeFFFF/Alt-F4/tree/master/assets/ALTF4). I think that is a remnant from when we didn't store images for the articles off of Github, but now it serves little purpose. 

It's a tiny annoyance each week, not sure how much work it would be to change though. The solution I have in mind is to have it just be one folder and we'd give each thumbnail a unique name that includes the issue number.

DrMagPie:
closes #540 

Added Default for thumbnail image.

if it is not specified in front matter Jekyll will put following url there: `/assets/ALTF4/{issue number}.jpg`


_posts/cs/2021-05-14-ALTF4-35.md
```diff
@@ -3,7 +3,6 @@ author: Romner, Conor_
 editor: stringweasel, Nanogamer7, Therenas, Firerazer
 translator: oof2win2
 image:
```
Conor12345, line 5: test

PR #545, opened 2021-08-18
Optimizing the solution

dcbroad3:
### Author info
Name: ilbJanissary
Suggested release date: Any

Discord: ilbJanissary#9596


submissions/optimizing_the_solution/optimizing-the-solution.md
```diff
@@ -0,0 +1,281 @@
...
+
+While we do not necessarily have anything we want to maximize in our linear program, we do want to minimize the amount of resources and factories required to produce our output items. So, instead of adding an objective we want to maximize, we can first add a "Cost" function to our linear program so that we can describe how we want to minimize the number of resources and factories in our solution.
+
+The cost function can be rather abstract and arbitrary, as there is no specific cost of resources and factories in games like Factorio and Dyson Sphere Program. Instead, we have to rely on intuition of what is most difficult to obtain and set up in a factory. In FactorioLab, this comes down to a few specific variables that define the cost function:
```
heinwessels: I think this paragraph could explain a little more. It's a weird concept that you add your own costs to the different items. It's obviously required, but new readers/players won't realize that there isn't enough knowns to simply solve the equation. Maybe you could give an example, or explain that without adding your own costs will results in infitite solutions. (I think, my math is rusty :P )

heinwessels (changes requested):
Hello! Wow, this is an awesome submission! Thanks for sharing this with Alt-F4! It's thorough and detailed, so I don't have much feedback :P

- [x] I think it's imporant to add an conclusion to tell the reader why what he/she just read was important. You could hit some points like:
  - Most big calculator mods like Factory Planner and Hellmod also do this
  - It will help you build ratio-perfect fluid handling which you can't do otherwise
  - It's cool how you can use calculus to build very satisfying builds.
  - This is a nice tool to use while you're at work or away from your PC
  - These are just ideas, anything will be cool :P

I really enjoyed this read, and I'm sure other people will too :)

dcbroad3:
@heinwessels 

> I think it's imporant to add an conclusion to tell the reader why what he/she just read was important. 

Now that you mention it, this seems very obvious. 

> Most big calculator mods like Factory Planner and Hellmod also do this

Actually, Factory Planner doesn't have a Simplex solver currently. It has a matrix solver, but it can only find a solution if the matrix is square. At least, as far as I understand it. 

So, it didn't feel right to just manufacture a **Conclusion** section at the bottom, it felt a bit too formal and forced, so I tried adding a **TL;DR** instead. Let me know if that's not exactly the kind of thing you were looking for.

submissions/optimizing_the_solution/optimizing-the-solution.md
dcbroad3: Tried splitting this into two sections, **Why do we need a cost function?** and **How do we define a cost function?**. Kirk McDonald's original essay does a bit of a better job in this area of showing graphically the line of possible solutions and highlighting that one point along the line is the optimal solution. Hopefully this helps make it more clear though.

heinwessels:
@dcbroad3 it looks great! And the conclusion is perfect. 

I had one final thought, do you want to add a snapshot of your calculator showing the output of a simple recipe? You're UI is quite nice looking. I was thinking just above the `Why use a calculator?` or `Concepts` heading.

dcbroad3:
> I had one final thought, do you want to add a snapshot of your calculator showing the output of a simple recipe? You're UI is quite nice looking. I was thinking just above the `Why use a calculator?` or `Concepts` heading.

Added, let me know if you have any other ideas 😄

Nanogamer7:
**Fact Check:** 

Since you are presenting your own work, there isn't that much to fact check, however I still listed the few minor details I found below

- ✔ All recipes are correct, however it shou 
- ✔ All links direct to where they say they are going, but maybe consider adding a link for:
  - 

---

Further notes:

---

✔ Correct
❗ Incorrect

Nanogamer7:
**Fact Check:** 

Since you are presenting your own work, there isn't that much to fact check, however I still listed the few minor details I found below

- ✔ All recipes are correct.
  - ❗ It should be noted, however, that actual crafting time might differ from recipe times due to differing crafting speeds. (i.e. "... by hand or by a mining drill." is not 100% accurate)
  - ❔ Technically speaking (read: "Not directly relevant for your article"), raw resources are separate to recipes in the game code [[1](https://wiki.factorio.com/Data.raw#recipe-category)][[2](https://wiki.factorio.com/Data.raw#resource-category)]. Furthermore water isn't considered a raw resource [[3](https://wiki.factorio.com/Data.raw#resource)], thus the pumpjack is presumably considered a type of miner, while the offshore pump presumably just generates water out of nothing, a shoreline merely being a requirement to place it. Also the offshore pump requires units of fluids per tick [[4](https://wiki.factorio.com/Prototype/OffshorePump#pumping_speed)], and thus the *recipes* (recipe in the context you are using it for FactorioLab) would technically be 1/60 second = 20 water.
- ✔ Your articles logic appears to be consistent
  - ❗ https://arxiv.org/abs/1810.07896 had a very long loading time the first time

---

Further notes:
- I really appreciate the use of icons, they make some parts way easier to read

---

✔ Correct
❗ Incorrect

dcbroad3:
@heinwessels I added images for recipes across the document, let me know if you have any concerns. 

@Nanogamer7 I added a couple notes about crafting speed, this article kind of lucks out because the real calculations are done mostly with factories where crafting speed is 1 and can be ignored but I agree it should be mentioned at least. I'm not sure that it's worthwhile to go into the technical details on what amounts to a raw resource and getting the recipe right at the tick level, especially since for the sake of the calculator it really doesn't care. I could easily update the water recipe to be correct if we wanted it to use the 1/60 second version of the recipe, but I thought it'd be more clear to the average reader as 1200/s.

Nanogamer7:
Your addition definitely resolves my point, and yes, the technical details are definitely not relevant for your article, I just had some fun digging through them ;)

PR #560, opened 2021-09-24
Fix atom feed format

9p4:
This PR fixes the atom feed format

### DIFF
--- a/atom.xml
+++ b/atom.xml
@@ -3,10 +3,10 @@ layout: null
 ---
 
 <?xml version="1.0" encoding="utf-8"?>
-<feed xmlns="https://www.w3.org/2005/Atom">
+<feed xmlns="http://www.w3.org/2005/Atom">
 
   <title>{{ site.title }}</title>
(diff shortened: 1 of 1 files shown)


justchen1369:
Is switching to http:// necessary?

9p4:
Yes, since it is part of the spec to specify it as http. This XML feed is invalid without it.

9p4:
Spec: https://validator.w3.org/feed/docs/atom.html
Failed Validation: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Falt-f4.blog%2Fatom.xml

justchen1369:
LGTM then 👍

9p4:
Seems it was introduced in this mass change. About the time ALT-F4 stopped working in my RSS aggregator.

https://github.com/AlternativeFFFF/Alt-F4/commit/8442b99628251e23501ca6b0f3552eaa51766f4c

Some other files (svg, etc) may need to be switched back as well (but I doubt there is much of an issue there).

PR #567, opened 2021-10-25
The Fluidic Power Experiment

heinwessels:
Name: stringweasel
Data: Maybe 5 November?


heinwessels:
Now split in two. Continued in #576

PR #586, opened 2021-11-13
Laziness through automation

ClaudeMetz:
### Author info
Name: Therenas
Suggested release date: whenever

Still missing links and a video, will  get to those soon-ish.


_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+Compounding this effect and creating a bit of a virtuous cycle is human psychology. If you know that your screenshots will need to be updated every second release, you'll just decide to skip updating them, only this once. The next release will have another change anyways, so why bother updating them now, that would be a waste of time. Only when the next release comes, the same psychology will lead you to procrastinate once more. Just this once. Repeat ad infinitum.
+
+Maybe I'm just incredibly lazy and normal people don't have any issues with this. They'd probably just update their screenshots on a regular schedule, and thus they'll be mostly up to date most of the time. But not me. In practice, it lead to me updating the screenshots about once every six months, if that. And it bothered me that they were not up to date, but the 'I'll just do it next time'-mentality was overpowering in the face of doing work that would be naught a week later.
```
MyNameIsTrez:
```suggestion
Maybe I'm just incredibly lazy and normal people don't have any issues with this. They'd probably just update their screenshots on a regular schedule, and thus they'll be pretty up-to-date most of the time. But I find this hard to keep up. In practice, it lead to me updating the screenshots about once every six months, if that. And it bothered me that they were not up-to-date, but the "I'll just do it next time"-mentality was overpowering in the face of doing work that would be naught a week later.
```
I think the original `But not me.` isn't a proper follow-up to the sentence before it, as the sentence before it ends with `and thus they'll be pretty up-to-date most of the time`, even though I get what you meant. I'm not totally sure what I'd prefer it to be replaced with, but something akin to the change I suggested is what I had in mind.

I replaced the single quotes with double quotes for the sake of making it clearer that the apostrophe in the `I'll` isn't the ending of the opening apostrophe before it.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+Still though, I wanted a solution. I needed a solution, if only to absolve me of my self-imposed guilt of having outdated screenshots. It couldn't be difficult to do of course. The idea of working on myself as a person would not pass muster, I'll do that next week. Automation was the promised land it seemed, as it often does, and as it sometimes is.
```
MyNameIsTrez, line 13:
```suggestion
Still though, I wanted a solution. I needed a solution, if only to absolve me of my self-imposed guilt of having outdated screenshots. It couldn't be difficult to find such a solution, of course. The idea of working on myself as a person would not pass muster, I'll do that next week. Automation was the promised land it seemed, as it often does, and as it sometimes is.
```
I changed the `It couldn't be difficult to do of course` to `It shouldn't be difficult to find such a solution, of course` because I think the sentence could be interpreted both as meaning that it shouldn't be difficult to find a solution, but also that it *literally couldn't* be hard to find a solution. But this may just be me rambling here, haha. The comma before the "of course" definitely should be added, though.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+I automated a few processes related to mod development before, most notably the process of [bundling up](https://github.com/ClaudeMetz/FactorioScripts/blob/5aab7569acdf86ff65167584638a3dd7323d2d0b/build_release.py) a new release and all the micro-management that comes with that. In that particular case, it's not even really because I didn't want to do the same tedious steps for all of my releases ([169 and counting](https://mods.factorio.com/mod/factoryplanner/downloads) for Factory Planner), it's more about preventing me from making mistakes during that process, which could lead to broken versions getting to users and so on, which would be bad.
```
MyNameIsTrez:
```suggestion
I automated a few processes related to mod development before, most notably the process of [bundling up](https://github.com/ClaudeMetz/FactorioScripts/blob/5aab7569acdf86ff65167584638a3dd7323d2d0b/build_release.py) a new release and all the micro-management that comes with that. In that particular case, it's not even really because I didn't want to do the same tedious steps for all of my releases ([169 and counting](https://mods.factorio.com/mod/factoryplanner/downloads) for Factory Planner); it's more about preventing me from making mistakes during that process, which could lead to broken versions getting to users and so on, which would be bad.
```

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Get to the Screenshotting Already!
+
+Alright alright, let's get to the fun part. My idea was pretty simple: Write a [Python script](https://github.com/ClaudeMetz/FactorioScripts/blob/5aab7569acdf86ff65167584638a3dd7323d2d0b/take_screenshots.py) that launches Factorio into a [custom scenario](https://github.com/ClaudeMetz/FactoryPlanner/tree/master/scenarios/screenshotter), which then takes over the actual screenshotting part (for which Factorio has a [dedicated API call](https://lua-api.factorio.com/latest/LuaGameScript.html#LuaGameScript.take_screenshot), thankfully), after which the script puts the screenshots into the right folder and maybe crops them a bit. The first part of that is rather easy, I just needed to figure out which [command line arguments](https://wiki.factorio.com/Command_line_parameters) I needed, string them together, and the scenario was up and running. So far, so simple.
```
MyNameIsTrez:
```suggestion
Alright, alright, let's get to the fun part. My idea was pretty simple: Write a [Python script](https://github.com/ClaudeMetz/FactorioScripts/blob/5aab7569acdf86ff65167584638a3dd7323d2d0b/take_screenshots.py) that launches Factorio into a [custom scenario](https://github.com/ClaudeMetz/FactoryPlanner/tree/master/scenarios/screenshotter), which then takes over the actual screenshotting part (for which Factorio has a [dedicated API call](https://lua-api.factorio.com/latest/LuaGameScript.html#LuaGameScript.take_screenshot), thankfully), after which the script puts the screenshots into the right folder and maybe crops them a bit. The first part of that is rather easy: I just needed to figure out which [command line arguments](https://wiki.factorio.com/Command_line_parameters) I needed, to string them together, and the scenario would already be up and running. So far, so good.
```
I'm not sure what to replace `string them together` with, but it feels like the wrong tense to me, or something like that. I added `to` in front of it, but maybe that shouldn't be added. ¯\\_\_(ツ)\_\_/¯

I think `So far, so simple` isn't a saying so I replaced that with `So far, so good`. You may want to either remove or keep the comma in `So far, so good`, as both get dictionary results on Google.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The actual scenario is a bit more complicated though. To be able to take meaningful screenshots, I need to setup various GUIs in various states to show the functionality I want to highlight. Doing this kind of thing for entities on the map is easier, since you can use all the APIs for interacting with the game world that you'd normally use. For GUIs though, which are a bit of a world on their own, you'll have to come up with your own system.
```
MyNameIsTrez:
```suggestion
The actual scenario is a bit more complicated, though. To be able to take meaningful screenshots, I need to set up various GUIs in various states to show the functionality I want to highlight. Doing this kind of thing for entities on the map is easier, since you can use all the APIs for interacting with the game world that you'd normally use. For GUIs though, which are a bit of a world on their own, you'll have to come up with your own system.
```
`to show the functionality I want to highlight` is perfectly valid probably, but `show` and `highlight` may be synonymous enough where you'd want to rephrase this.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The biggest issue comes from not being able to just simulate an actual user clicking on your interface. The game does not have any API method for this, since it's not needed for any actual mod. What it means though is that you need to somehow emulate clicks. This is a bit awkward since, unless your mod is setup for this from the start, or just very well structured, you'll need to either rewrite your event handler or use rather hacky methods. You might be able to guess which route I took.
```
MyNameIsTrez:
```suggestion
The biggest issue comes from not being able to just simulate an actual user clicking on your interface. The game does not have any API method for this, since it's not needed for any actual mod. What it means, though, is that you'll need to somehow emulate clicks. This is a bit awkward since, unless your mod is set up for this from the start, or just very well-structured, you'll need to either rewrite your event handler or use rather hacky methods. You might be able to guess which route I took.
```

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+After getting the [hacks](https://github.com/ClaudeMetz/FactoryPlanner/blob/master/modfiles/data/handlers/screenshotter.lua) sorted, this whole contraption went as far as launching the game, setting up the interface into an interesting scenario, and taking screenshots of various parts of the UI. I have to point out that all of this is very duct-tape-y and will likely break very easily as the mod code changes. In this particular case though, there's not really an alternative, so I'll have to put up with it. It's much more fun anyways to build a giant pile of spaghetti code than manually taking the same screenshots over and over, so it's a win either way.
```
MyNameIsTrez:
```suggestion
After getting the [hacks](https://github.com/ClaudeMetz/FactoryPlanner/blob/master/modfiles/data/handlers/screenshotter.lua) sorted, this whole contraption went as far as launching the game, setting up the interface into an interesting scenario, and taking screenshots of various parts of the UI. I have to point out that all of this is very duct-tape-y and will likely break very easily as the mod's code changes. In this particular case, though, there's not really an alternative, so I'll have to put up with it. It's much more fun to build a giant pile of spaghetti code than manually taking the same screenshots over and over anyways, so it's a win either way.
```
I assume `as the mod code changes` refers to the mod you're writing, and not the code that runs mods, aka the Factorio engine? I've gone with the former for now, but change this if you meant the latter.

`In this particular case, though, there's not really an alternative, so I'll have to put up with it.`
This contradicts your `you'll need to either rewrite your event handler or use rather hacky methods` statement, so I'd either remove this contradiction entirely or reword it.
`an alternative` may also sound better when exaggerated to `any alternative`

I moved the `anyways` from `It's much more fun anyways` to the end of the sentence, but feel free to revert that if you prefer keeping it like it was or to move it to be `a giant pile of spaghetti code anyways than`

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+I was pretty happy with what I had done so far, but still, maximum automation this was not. See, there was still the matter of cropping the screenshots to a smaller size. A lot of the dialogs in Factory Planner are relatively small, and the having game world in the background would be distracting. I could do the cropping part manually, but I didn't want to. *Something something* lazyness. I had a potential solution in the back of my mind, but it would require some dirty work.
```
MyNameIsTrez:
```suggestion
I was pretty happy with what I had done so far, but still, maximum automation this was not. See, there was still the matter of cropping the screenshots to a smaller size. A lot of the dialogs in Factory Planner are relatively small, and having the game world in the background would be distracting. I could do the cropping part manually, but I didn't want to. *Something something* laziness. I had a potential solution in the back of my mind, but it would require some dirty work.
```
> maximum automation this was not.
I'm not sure if this intentionally sounds theater/Yoda-ish, but I just wanted to highlight this part in case it wasn't intentional. Feel free to leave it.

> but it would require some dirty work.
You may want to change this to `dirty-work`, in case you meant it as a noun instead of a verb. Feel free to leave it.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The basic idea was to, as the scenario was taking the screenshots, also record the sizes of the dialogs whose picture was being taken. This metadata could then be written to a file as well, and the Python script could use it to crop the screenshots to the right size. This sounds like a great idea in theory, but it runs into a tiny issue in reality: The game doesn't let you read the dimensions of GUI windows, only write them.
```
MyNameIsTrez:
```suggestion
As the scenario was taking the screenshots, the basic idea was to also record the dimensions of the dialogs whose pictures were being taken. This metadata could then be written to a file as well, and the Python script could use it to crop the screenshots to the right size. This sounds like a great idea in theory, but it runs into a tiny issue in reality: The game doesn't let you read the dimensions of GUI windows, only write them.
```
I originally wanted to change to `The basic idea was to—as the scenario was taking the screenshots—also record the sizes of the dialogs whose picture was being taken.` as em-dashes are used for interjections ("Interject or abruptly change the subject" on [Grammarly](https://www.grammarly.com/blog/semicolon-vs-colon-vs-dash/); scroll down to the "Em dash" header to see examples). I think the interjection at the very start doesn't flow well, though, as you haven't painted what the basic idea is at this point. I've restructured it to what I think flows better, instead of adding the em-dashes.

I've changed `to also record the sizes of the dialogs` to `to also record the dimensions of the dialogs` to be consistent with your `The game doesn't let you read the dimensions of GUI windows, only write them.` sentence.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+This is forbidden for good reason, as these dimensions are not actually deterministic in all cases, as a different game language for example might change the size of certain elements. Reading and using these undeterministic values in multiplayer would lead to desyncs, making the mod multiplayer-incompatible, and break replays too. Seemed like a dead end. Was I fated to forever manually crop screenshots? Turns out, I was not.
```
MyNameIsTrez:
```suggestion
This is forbidden for a good reason, as these dimensions are actually not deterministic in all cases; setting the game to a different language might change the sizes of certain elements, for example. Reading and using these non-deterministic values in multiplayer would lead to desyncs, which would make the mod multiplayer-incompatible, and it'd also break replays. This seemed like a dead end. Was I fated to manually crop screenshots forever? Turns out, I was not.
```

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+Now, I couldn't solve the issue that reading frame dimensions leads to desyncs, that's just how the game works on a fundamental level. I am however not running the game in multiplayer to take these screenshots, and as it turns out, you can already launch the game in a special mode called [Instrument Mode](https://lua-api.factorio.com/latest/Instrument.html) that disables multiplayer and enables some unsafe functionality. Seems like the perfect place to slip in a super-secret hidden method that gives me access to the forbidden knowledge of frame dimensions. So I went ahead and put it in. Working for Wube does have its perks as it turns out.
```
MyNameIsTrez:
```suggestion
Now, I couldn't solve the issue that reading frame dimensions leads to desyncs, that's just how the game works on a fundamental level. I am, however, not running the game in multiplayer to take these screenshots, and as it turns out, you can already launch the game in a special mode called [Instrument Mode](https://lua-api.factorio.com/latest/Instrument.html) that disables multiplayer and enables some unsafe functionality. This seems like the perfect place to slip in a super-secret hidden method that gives me access to the forbidden knowledge of frame dimensions. So I went ahead and put it in. Working for Wube does have its perks, as it turns out.
```

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+After several days of piling on the hacks, the beast finally roared to life, fully automating the process of taking screenshots for me. Although, *fully* automating the process is not quite correct. Turns out, it's still not fully automated. Of course it's not. Why would I get to be *fully* satisfied. There's two steps that I still need to do myself.
```
MyNameIsTrez:
```suggestion
After several days of piling on the hacks, the beast finally roared to life, fully automating the process of taking screenshots for me. Although *fully* automating the process is not quite correct. Turns out, it's still not fully automated. Of course it's not. Why would I get to be *fully* satisfied. There's two steps that I still need to do myself.
```

MyNameIsTrez: I don't think `Turns out` should be replaced with `It turns out` here, but it depends on the balance of casual/formal you're aiming for. Although I've been changing it to be more formal in the previous suggested changes, you might want to make it consistently either casual or formal everywhere.

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The first is quitting the game after the screenshots are taken, so that the Python script can continue. As it turns out, a mod can not make Factorio quit to desktop. That's entirely reasonable to prevent abuse, even though it would be a nice option to have for Instrument Mode. Either way, hitting Alt-F4 (*ding*) at the right time is something I can manage I think. One intrusive thought I had was to find a way to make the game crash to achieve my goal. I warned you that this is all a pile of hacks. The issue with that plan is that there's very few of these bugs in the game, and once they are found, you can be sure the bug-swatter will squash them immediately.
```
MyNameIsTrez:
```suggestion
The first is quitting the game after the screenshots are taken, so that the Python script can continue. As it turns out, a mod can not make Factorio quit to desktop. That's entirely reasonable to prevent abuse, even though it would be a nice option to have for Instrument Mode. Either way, hitting Alt-F4 (*ding*) at the right time is something I can manage manually, I think. One intrusive thought I had was finding a way to make the game crash to quit to desktop. I warned you that this is all a pile of hacks. The issue with that plan is that there's very few of these bugs in the game, and once they are found, you can be sure the bug-swatter will squash them immediately.
```
I *think* you're talking about manually pressing Alt-F4 here, but you could also let Python do this for you automatically, so ¯\\_\_(ツ)\_\_/¯

I personally also like reiterating what was discussed moments before, so that's why I replaced `achieve my goal` with `quit to desktop`

_drafts/screenshotter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The second is uploading these hot and steamy new screenshots to the mod portal. There is no proper API for manipulating the mod portal listings at the moment, so I'm out of luck there. For now, I'll have to manually open a browser, delete my screenshots one-by-one, and upload the new ones one-by-one. *Quelle tragédie!*. Maybe there is hope for an API like this in the future though, we'll see.
+
+Besides those two things though, I'm pretty happy with how this turned out. I think the juice was worth the squeeze, as I get to harvest the benefits of this script indefinitely. The joys of automation: *Infinite gains*. At least until the first time it breaks, which is likely very soon. Oh well.
```
MyNameIsTrez:
```suggestion
Besides those two things though, I'm pretty happy with how this turned out. I think the juice was worth the squeeze, as I get to harvest the benefits of this script indefinitely. The joys of automation: *Infinite gains*. At least until the first time it breaks, which is likely very soon. [Oh well](https://xkcd.com/1319/).
```
I had to.

MyNameIsTrez (approved):
Great job! My stuff is very nitpicky as your article was already pretty much done!

PR #624, opened 2022-04-15
Fixing behaviour of 404 page

### LINKED ISSUE #568: [minor] Language popout is not showing selected language
The language popout is showing English even though I switched it to French (just as an example).
Screenshot 2021-10-26 at 16 58 17

This however does not occur on the article pages.
Screenshot 2021-10-26 at 16 59 42

DrMagPie:
fixes #568

Cleaned up and remove unnecessary hacks from  404 page.


DrMagPie:
hmm interesting, for some reason French is broken....
but I can't find a reason.
Locality it builds successfully and French is available.
but on netlify it's broken. nothing useful in logs.

DrMagPie:
okay, the classic "turn it off and on again" helped.
have no idea what was causing the problem...

PR #628, opened 2022-05-08
factorio-draftsman: A Python module for creating blueprint strings

redruin1:
### Author info
Name: redruin1
Suggested release date: 

- [ ] Images are hosted on Media server 
- [ ] Gifs are hosted on Media server
  - [ ] Gifs are [compressed](https://www.iloveimg.com/compress-image/compress-gif) 
- [ ] Videos are hosted on Youtube or Media Server
  - [ ] Make sure video is public or unlisted
  - [ ] Use provided Youtube embed ``
- [ ] Audio is hosted on souldcloud
  - [ ] Embed soundcloud post of audio file


submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+from draftsman.entity import ConstantCombinator
+
+
+def main():
+    blueprint = Blueprint()
+
+    count = 0  # Total number of signals
+    index = 0  # Signal index in the current combinator
+    i = 0  # How many combinators we've added
+    x = 0
```
MyNameIsTrez, line 86: Screenshot_20220508-135655~2
The `x = 0` and `y = 0` in your code can be removed, as seen here.

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+        combinator.set_signal(index, item, stack_size)
+        index += 1
+        # Once we exceed the current combinator, place it and reset
+        if index == 20:
+            # Add the combinator to the blueprint
+            combinator.id = "{}_{}".format(x, y)
+            blueprint.entities.append(combinator)
+            # Reset the combinator
+            i += 1
+            y = i % 5
```
MyNameIsTrez: I recommend adding a constant for that 5 where you've commented at the top of the file that you're going to create an Nx5 grid of connected constant combinators.

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+from draftsman.data import items
+from draftsman.entity import ConstantCombinator
+
+
+def main():
+    blueprint = Blueprint()
+
+    count = 0  # Total number of signals
+    index = 0  # Signal index in the current combinator
+    i = 0  # How many combinators we've added
```
MyNameIsTrez: Personally I suggest removing the comments and changing `count` to `signal_count`, `index` to `signal_index` and `i` to `combinators_added`.

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+            if "hidden" in items.raw[item]["flags"]:
+                continue
+        # Keep track of how many signals we've gone through
+        count += 1
+        # Write the stack size signal
+        stack_size = items.raw[item]["stack_size"]
+        combinator.set_signal(index, item, stack_size)
+        index += 1
+        # Once we exceed the current combinator, place it and reset
+        if index == 20:
```
MyNameIsTrez: Can you add a constant for 20 in case there is some significance to the number? I presume it is how many signals can be defined in a constant combinator, but I haven't really used combinators much. `signals_per_combinator` or `signals_per_constant_combinator`?

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+    combinator.id = "{}_{}".format(x, y)
+    blueprint.entities.append(combinator)
+
+    # Add connections to each neighbour
+    for cx in range(x):
+        for cy in range(5):
+            here = "{}_{}".format(cx, cy)
+            right = "{}_{}".format(cx + 1, cy)
+            below = "{}_{}".format(cx, cy + 1)
+            try:
```
MyNameIsTrez: You need the try-excepts because you're always connecting columns of combinators at a time without taking into account that the last combinator isn't necessarily at the bottom of its column, right? I recommend using this instead, since it is clearer on when an attempted connection would have a `KeyError` and since it immediately breaks instead of continuing to loop (only breaks out of the inner loop, but that's fine since `cx == x` means the outer loop is done anyways):
```py
if cx == x && cy > y: break
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
+# Introduction
+
+A couple months ago, I decided that I wanted to try my hand at making a **Self-expanding factory** in Factorio. After seeing an number of [impressive](https://www.youtube.com/watch?v=xF--1XdcOeM) [examples](https://www.youtube.com/watch?v=KNSBerFe_H8), I was inspired to have a go at the problem. I already had an outline of the logic, and how the factory would keep track of itself, as well as lofty ideas on impressive things that I could make it do. The only trouble was that I had actually never used combinators before, and I was planning on using them for the actual decision-making.
```
heinwessels: Maybe instead of referencing the JOSEF video you can reference the article about it :)
https://alt-f4.blog/ALTF4-39/

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+
+That's no problem though, we just create a scrap world, switch to the editor and start playing around!
+
+![death](./img/map_view_of_madness.png)
+
+![additional death](./img/madness.png "Any attempt to understand the meaningless scribbles of a madman will only result in disappointment and despair.")
+
+![too much playtime](./img/play_time.png "ROOKIE NUMBERS")
```
heinwessels: I think we can turn this into a ["gallery"](https://alt-f4.blog/ALTF4-X/#gallery) so that the images don't take to much space proportially. All you have to do is add a little comment and we can add it later.
```suggestion
// Conor, these three images should be a gallery
![death](./img/map_view_of_madness.png)

![additional death](./img/madness.png "Any attempt to understand the meaningless scribbles of a madman will only result in disappointment and despair.")

![too much playtime](./img/play_time.png "ROOKIE NUMBERS")
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+* **Draftsman is close-to-source.** Draftsman bases all of its data off of [Wube's `factorio-data` repository](https://github.com/wube/factorio-data), which means that all entities are exactly as you would expect in game with no inconsistencies. This makes Draftsman up-to-date, makes updating Draftsman for future Factorio versions a breeze, and allows versioning control to monitor changes between Draftsman and Factorio if any future breakage occurs.
+* **Draftsman supports mods.** Draftsman emulates the [Factorio data lifecycle](https://lua-api.factorio.com/latest/Data-Lifecycle.html) directly, meaning that the same loading process that happens when you launch the game is mimicked with a single Draftsman function. In addition to ensuring absolute accuracy to Factorio, this also makes it compatible with custom mod prototypes that can be accessed in script as if they were any other vanilla entity.
+
+Draftsman also has support for custom `"EntityLike"` objects, most notably `Group` objects, that allow you to create custom constructs that can be inserted into blueprints for aid in clarity and compartmentalization. Again, *MAYBE* I got a little carried away.
```
heinwessels: I think this paragraph can use an example of what you can group together.

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+
+In an effort to keep this document brief, I'm not going to go into the nitty-gritty about the module or how exactly it works; I've spent a lot of time writing documentation for that purpose instead. Rather, I'm going to show off a couple of things that I've made with it so far, as well as potential things that could be made in the future, to try to illustrate exactly why I spent all this time in the first place.
```
heinwessels:
```suggestion
In an effort to keep this article brief, I'm not going to go into the nitty-gritty about the module or how exactly it works; I've spent a lot of time writing documentation for that purpose instead. Rather, I'm going to show off a couple of things that I've made with it so far, as well as potential things that could be made in the future, to try to illustrate exactly why I spent all this time in the first place. 
```
`document` sounds so formal :P

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+
+This script is concise and easy to follow, but the really amazing thing about Draftsman is that this script is completely dynamic and can adjust to whatever version of Factorio that you're playing. New items, removed items, or changed stack sizes, regardless of the author of the change, are resolved automatically with the exact same code. For illustration, pictured top-left is what's generated by running the script for a vanilla game, top-right a game with the medium-sized overhaul modpack [Space Exploration](https://mods.factorio.com/mod/space-exploration), and at the bottom a [Bobs](https://mods.factorio.com/user/Bobingabout) + [Pyanodon](https://mods.factorio.com/user/pyanodon) mega-modpack:
+
+![Results](./img/item_stack_size_output.png)
+
+## Image to Blueprint Converter
+
+[This was something I made on a whim.](https://github.com/redruin1/factorio-draftsman/blob/main/examples/image_converter.py) It uses Pillow to load an image and converts it to a blueprint intended to be visible from the map view, all in less than 150 lines of code:
+
+![Portrait of Mike](./img/mike_stoklasa.png "A drunk bum with comically large flask, colorized.")
```
heinwessels: Personally I find the description funny, but I think we need to make it a bit more family friendly :)

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,200 @@
...
+* Dithering is only implemented on 1x1 tiles; I was having trouble adjusting the algorithm to multi-tile entities.
+* Multi-tile entities also fail to adjust the error metric property; palettization doesn't work as well when some pixels are of different size.
+* Some multi-tile entities have different rotations; a better implementation would check which orientation produces the least error before placing, instead of only placing them with the default orientation.
+* The colors themselves are hardcoded; it would be nice to dynamically load this from the game, especially with modded map colors...
+
+Despite these shortcomings, the output actually ends up pretty good for it's quick and dirty implementation. This shows that Draftsman is versatile enough for some rather eclectic purposes.
+
+![Portrait of Jay](./img/jay_bauman.png "Jay Bauman, notorious hack fraud, criminal at large. Armed and dangerous.")
+
+## Factorio Movie Player Resurrection
```
heinwessels: Ah man this is epic!

heinwessels (commented):
Hello! Thanks for your submission! It's really well written, you're a good story teller. I don't have much to add, except for a minor thing here and there. Really cool project too!

submissions/factorio_draftsman/main.md
redruin1: Added as global constant `COMBINATOR_HEIGHT`.

submissions/factorio_draftsman/main.md
redruin1: Done.

submissions/factorio_draftsman/main.md
redruin1: Changed to:
```python
if signal_index == combinator.item_slot_count:
    # ...
```
The 20 was leftover from when the item_slot_count of constant combinators wasn't extracted. Now it can be used and is dynamic to whichever combinator was specified in the beginning.

submissions/factorio_draftsman/main.md
redruin1: Well, more so that each combinator is supposed to be connected to its right and below neighbor in order to create a regular grid of wires. However, if it's on the rightmost column, it will attempt to connect right to an entity that's not there, and on the bottom row, connect below to an entity that's not there, hence the `KeyError`s. `KeyError` in this case could be renamed to `MissingEntityError`, which would be more specific. I could iterate by `x - 1`, by `5 - 1` instead, but that wouldn't complete all the connections on the right column or bottom row, because half of the connections on those rows are valid. Instead of messing around with complex conditions or multiple loop passes, I found the simplest way to connect them was to simply iterate over a grid of the same size and use Python's `try` to attempt the connection if possible and `pass` (ignore the error) if it happens to fails on the edges.

I admit though that this is ugly, and to this day I've never been a fan of the way it looks. I could do a similar method without of checking using the try catch, and explicitly check to see if the entity exists before attempting the connection; would this be clearer?
```python
for cx in range(x):
    for cy in range(COMBINATOR_HEIGHT):
        here = "{}_{}".format(cx, cy)
        right = "{}_{}".format(cx + 1, cy)
        below = "{}_{}".format(cx, cy + 1)

        # get() in this case would attempt to get the value of key right, and return False if not found
        if blueprint.entities.get(right, False): 
            blueprint.add_circuit_connection("red", here, right)
        if blueprint.entities.get(below, False):
            blueprint.add_circuit_connection("red", here, below)
```
*Technically* this isn't implemented yet, so I guess this is also an API question. `get()` is commonly used on dictionary-like objects, of which my `EntityList` has functionality from. Alternatively, you could also have something like `blueprint.contains(string_id)`.

All other of your proposed changes have been resolved.

submissions/factorio_draftsman/main.md
redruin1: Fixed.

submissions/factorio_draftsman/main.md
redruin1: Resolved to `A peculiar man with a comically large flask, colorized.` Let me know if the alcohol is a problem as well.

submissions/factorio_draftsman/main.md
redruin1: Added an example paragraph between the two.

submissions/factorio_draftsman/main.md
redruin1: Done. Forgot that you guys already had an article on it.

submissions/factorio_draftsman/main.md
redruin1: https://www.youtube.com/watch?v=jrHA6JlrOHE

submissions/factorio_draftsman/main.md
heinwessels: Nah, I think that's perfect. Who knows, could be a coffee flash :P Thanks

submissions/factorio_draftsman/main.md
MyNameIsTrez: I see, I had the wrong idea of what the loop was doing, thank you for elaborating!

Keep the try-excepts you have now then, Ask Forgiveness Not Permission and all being the Python mantra for clean code. :)

This stackoverflow answer on AFNP even uses your case of dictionaries as an example!
https://stackoverflow.com/a/12265860/13279557

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,212 @@
...
+
+* Many implementations were highly specific to the domain they were written for; a combinator computer compiler script was not going to be portable for many other purposes than compiling code for a specific combinator computer.
+* They lacked a unified language; many were written in Lua, some in Python, another in JavaScript, one in C++, etc. This meant that each one had to write it's own implementations for the same operations, instead of just having one guy write the implementation they need and making it available to others.
+* Many of them were written for versions of Factorio now severely out of date.
+* Documentation for a lot of these modules was sparse and sporadic, which turns off users like me who want to know what the module is capable of before investing any time learning how to use it.
+
+Fundamentally unsatisfied with the options, I resigned myself to my fate. I was going to have to follow in the footsteps of all before me, and develop my own implementation from scratch. How arduous!
+
+I made a prototype script in an afternoon that did exactly what I needed and it worked perfectly. It used a template system and it took less than a week to completely finish.
```
LevFendi: Less than a week to finish your combinator computer, right?

LevFendi (commented):
Thank you for considering my previous suggestions, I think the article is much clearer now and it is an excellent article!

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
+## Draftsman <author>redruin1</author>
+
+A couple months ago, I decided that I wanted to try my hand at making a **self-expanding factory** in Factorio. After seeing an number of [impressive](https://www.youtube.com/watch?v=xF--1XdcOeM) [examples](https://alt-f4.blog/ALTF4-39/), I was inspired to have a go at the problem. I already had an outline of the logic, and how the factory would keep track of itself, as well as lofty ideas on impressive things that I could make it do. The only trouble was that I had actually never used combinators before, and I was planning on using them for the actual decision-making.
```
MyNameIsTrez:
```suggestion
A couple months ago, I decided that I wanted to try my hand at making a **self-expanding factory** in Factorio. After seeing a number of [impressive](https://www.youtube.com/watch?v=xF--1XdcOeM) [examples](https://alt-f4.blog/ALTF4-39/), I was inspired to have a go at the problem. I already had an outline of the logic, and how the factory would keep track of itself, as well as lofty ideas on impressive things that I could make it do. The only trouble was that I had actually never used combinators before, and I was planning on using them for the actual decision-making.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+That's no problem though, we just create a scrap world, switch to the editor and start playing around!
```
MyNameIsTrez:
```suggestion
That's no problem, though: We just create a scrap world, switch to the editor and start playing around!
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+With all these examples, I was hoping that I could find some similarities with one of them and use that as a baseline for my solution to avoid "retreading" the same ground. Something troubled me about the solutions that were available though: They all had problems!
+
+* Many implementations were highly specific to the domain they were written for; a combinator computer compiler script was not going to be portable for many other purposes than compiling code for a specific combinator computer.
+* They lacked a unified language; many were written in Lua, some in Python, another in JavaScript, one in C++, etc. This meant that each one had to write its own implementations for the same operations, instead of just having someone write the implementation they need and making it available to others.
```
MyNameIsTrez:
```suggestion
* They lacked a unified language; many were written in Lua, some in Python, another in JavaScript, one in C++, etc. This meant that each one had to write their own implementations for the same operations, instead of just having someone write the implementation they need and making it available to others.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+Anyway, here's a Python module I made.
+
+### Introducing: `factorio-draftsman`
+
+![Draftsman logo, represented in-game, generated via script.](./img/draftsman_logo.png "I went through dozens of different combinations of tile orders before settling on this one.")
+
+[Draftsman](https://github.com/redruin1/factorio-draftsman) is a Python module for creating, modifying, importing, and exporting all manner of Factorio blueprint strings. The package allows you to create and design blueprints programmatically, to aide in the development of tedious and repetitive blueprints that would be untenable to create by hand, much like the problem I ran into described above. Draftsman attempts to solve all of the shortcomings of existing Factorio blueprint implementations:
```
MyNameIsTrez:
```suggestion
[Draftsman](https://github.com/redruin1/factorio-draftsman) is a Python module for creating, modifying, importing, and exporting all manner of Factorio blueprint strings. The package allows you to create and design blueprints programmatically, to aid in the development of tedious and repetitive blueprints that would be untenable to create by hand, much like the problem I ran into described above. Draftsman attempts to solve all of the shortcomings of existing Factorio blueprint implementations:
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+* **Draftsman does it all.** All types of entities are supported, from splitters to stack inserters. If you can do it in game, you can do it with Draftsman; allowing you to focus on the one problem you're actually trying to solve.
```
MyNameIsTrez:
```suggestion
* **Draftsman does it all.** All types of entities are supported, from splitters to stack inserters. If you can do it in-game, you can do it with Draftsman, allowing you to focus on the one problem you're *actually* trying to solve.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+* **Draftsman is uni-language.** Written in Python, this makes it exceptionally easy to install, simple to use, and gives the user access to the entirety of Python's vast packaging index. Chances are, you can do it in Python with Draftsman, regardless of *what* Factorio-related thing you're actually doing.
```
MyNameIsTrez:
```suggestion
* **Draftsman is uni-language.** Written in Python, this makes it exceptionally easy to install, simple to use, and gives the user access to the entirety of Python's vast packaging index. Chances are you can do it in Python with Draftsman, regardless of *what* Factorio-related thing you're actually doing.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+* **Draftsman is easy to use.** Designed from the start to be simple and (most importantly) self-documenting, Draftsman allows you to manipulate blueprints and entities by auto-completable attributes and methods.
+* **Draftsman is well-documented.** Every function, method, attribute, and class is documented and crosslinked at it's [readthedocs site](https://factorio-draftsman.readthedocs.io/en/latest/?badge=latest). In addition, tutorials and supplementary material is provided, as well as a whole host of [example programs](https://github.com/redruin1/factorio-draftsman/tree/main/examples) to help illustrate how Draftsman works.
```
MyNameIsTrez:
```suggestion
* **Draftsman is well-documented.** Every function, method, attribute, and class is documented and crosslinked at its [readthedocs site](https://factorio-draftsman.readthedocs.io/en/latest/?badge=latest). In addition, tutorials and supplementary material are provided, as well as a whole host of [example programs](https://github.com/redruin1/factorio-draftsman/tree/main/examples) to help illustrate how Draftsman works.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+![Portrait of Mike](./img/mike_stoklasa.png "A peculiar man with a comically large flask, colorized.")
+
+Many improvements could be made to this:
+
+* Dithering is only implemented on 1x1 tiles; I was having trouble adjusting the algorithm to multi-tile entities.
+* Multi-tile entities also fail to adjust the error metric property; palettization doesn't work as well when some pixels are of different size.
+* Some multi-tile entities have different rotations; a better implementation would check which orientation produces the least error before placing, instead of only placing them with the default orientation.
+* The colors themselves are hardcoded; it would be nice to dynamically load this from the game, especially with modded map colors...
+
+Despite these shortcomings, the output actually ends up pretty good for it's quick and dirty implementation. This shows that Draftsman is versatile enough for some rather eclectic purposes.
```
MyNameIsTrez:
```suggestion
Despite these shortcomings, the output actually ends up pretty good for its quick and dirty implementation. This shows that Draftsman is versatile enough for some rather eclectic purposes.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+* **Draftsman is stable.** A rigorous suite of tests ensure that Draftsman behaves predictably and correctly (or at least correctly enough), with a code coverage of 100%. Draftsman is verified to work on the latest versions of Python 2 and 3, and is compatible with all versions of Factorio greater than 1.0, with the metrics to prove it.
+* **Draftsman is descriptive.** Draftsman enforces "Factorio-safety" as a core philosophy, which means that if there is any change to a blueprint that would result in an error on import, an exception is thrown. Draftsman also tries to enforce "Factorio-correctness", which means that values that won't break, but are otherwise nonsensical, will raise warnings. Both errors and warnings are verbose, so that any problem with your script can be understood and resolved in seconds.
```
MyNameIsTrez: > which means that if there is any change to a blueprint that would result in an error on import, an exception is thrown

What is meant with "any change" here? Is this referring to the hypothetical of a future Factorio updated breaking some blueprints?

If not, I think replacing that part with this would be an improvement?:
> which means that if a blueprint would result in an error on import, an exception is thrown

Or maybe even specifying that `import` is referring to an action in Factorio, to make it clear it's not referring to giving a blueprint to Draftsman?:
> which means that if a blueprint would result in an error on import in Factorio, an exception is thrown

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+* **Draftsman is close-to-source.** Draftsman bases all of its data off of [Wube's `factorio-data` repository](https://github.com/wube/factorio-data), which means that all entities are exactly as you would expect in-game with no inconsistencies. This makes Draftsman up-to-date, makes updating Draftsman for future Factorio versions a breeze, and allows versioning control to monitor changes between Draftsman and Factorio if any future breakage should occur.
```
MyNameIsTrez:
```suggestion
* **Draftsman is close-to-source.** Draftsman bases all of its data off of [Wube's `factorio-data` repository](https://github.com/wube/factorio-data), which means that all entities are exactly as you would expect in-game, with no inconsistencies. This makes Draftsman up-to-date, makes updating Draftsman for future Factorio versions a breeze, and allows version control to monitor changes between Draftsman and Factorio, if any future breakage should occur.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+* **Draftsman supports mods.** Draftsman emulates [Factorio's data lifecycle](https://lua-api.factorio.com/latest/Data-Lifecycle.html) directly, meaning that the same loading process that happens when you launch the game is mimicked with a single Draftsman function. In addition to ensuring absolute accuracy to Factorio, this also makes it compatible with custom mod prototypes that can be accessed in script as if they were any other vanilla entity.
```
MyNameIsTrez: > that can be accessed in script

What do you mean by this? Factorio mods that use Lua? Draftsman that uses Python? Knowing this should also answer my question of how it'd be accessed as well.

submissions/factorio_draftsman/main.md
MyNameIsTrez:
```suggestion
* **Draftsman is descriptive.** Draftsman enforces "Factorio-safety" as a core philosophy, which means that if a blueprint would result in an error on import in Factorio, an exception is thrown by Draftsman. Draftsman also tries to enforce "Factorio-correctness", which means that values that won't break, but are otherwise nonsensical, will raise warnings. Both errors and warnings are verbose, so that any problem with your script can be understood and resolved in seconds.
```

submissions/factorio_draftsman/main.md
redruin1: Well, more the encoding script for that particular ROM. Mk1 of the computer had already been "finished" by this point, and that took several months. Should that be made more clear?

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+Draftsman has custom classes designed around each individual entity and prototype, and is designed to work as seamlessly as possible with blueprint strings and other software. You can import a blueprint string from Factorio and it will be automatically converted into a Draftsman `Blueprint` object with all its functionality, make any change you can think of to it, and then export that `Blueprint` object back into a string to be re-imported into Factorio. Or, you can just make an entirely new `Blueprint` from scratch; Draftsman is designed to work around you, not for you to work around it.
```
MyNameIsTrez:
```suggestion
Draftsman has custom classes designed around each individual entity and prototype, and is designed to work as seamlessly as possible with blueprint strings and other software. Draftsman converts to and from these prototypes when you import and export blueprint strings automatically, no extra steps required. You can import a blueprint string from Factorio as a `Blueprint` object, make any change you want, and then export that `Blueprint` object back into a string. Or, you can just make an entirely new `Blueprint` from scratch; Draftsman is designed to work around you, not for you to work around it.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+Draftsman also has support for custom `"EntityLike"` objects, most notably `Group` objects, that allow you to create custom constructs that can be inserted into blueprints for aid with clarity and compartmentalization. For example, you can make a design for a smelting block in a `Group`, and then you can place that block as many times as you want, where ever you want, rotated or flipped, essentially acting as a copy-paste tool within Draftsman.
```
MyNameIsTrez:
```suggestion
Draftsman also has support for custom `"EntityLike"` objects, most notably `Group` objects, that allow you to create custom constructs that can be inserted into blueprints for aid with clarity and compartmentalization. For example, you can make a design for a smelting block in a `Group`, and then you can place that block as many times as you want, wherever you want, rotated or flipped, essentially acting as a copy-paste tool within Draftsman.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+In an effort to keep this article brief, I'm not going to go too far into the nitty-gritty about the module or how exactly it works; I've spent a lot of time writing documentation for that purpose instead. Rather, I'm going to show off a couple of things that I've made with it so far, as well as potential things that could be made in the future, to try to illustrate exactly why I spent all this time making it in the first place.
+
+### Automatic Item Stack Sizes
+
+Oftentimes, for you want to figure out how much storage you need for a specific amount of an item. However, storage in Factorio is based on slots, not amounts, so the amount of storage is actually dependent on not only the amount you're trying to store, but also the stack size of that item. [You can design a circuit contraption](https://factorioprints.com/view/-LwH68dlvOEIKWkU5JqR) to determine this amount of slots that you need for any input item, but you'd need a big cell of combinators documenting the item sizes. This is *boring* to make, and easily broken if a new item is added or the stack size changes. Clearly, for such a simple and repetitive task, a script is better suited:
```
MyNameIsTrez:
```suggestion
Oftentimes, you want to figure out how much storage you need for a specific amount of an item. However, storage in Factorio is based on slots, not amounts, so the amount of storage is actually dependent on not only the amount you're trying to store, but also the stack size of that item. [You can design a circuit contraption](https://factorioprints.com/view/-LwH68dlvOEIKWkU5JqR) to determine the number of slots that you need for any input item, but you'd need a big cell of combinators documenting the item sizes. This is *boring* to make, and easily broken if a new item is added or the stack size is changed. Clearly, for such a simple and repetitive task, a script is better suited:
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+        # Once we exceed the number of signals a combinator can hold, place it and reset
+        if signal_index == combinator.item_slot_count:
+            # Add the combinator to the blueprint
+            combinator.id = "{}_{}".format(x, y)
+            blueprint.entities.append(combinator)
+            # Reset the combinator
+            combinators_added += 1
+            y = combinators_added % COMBINATOR_HEIGHT
+            x = int(combinators_added / COMBINATOR_HEIGHT)
+            combinator.set_signals(None)  # Clear signals
```
MyNameIsTrez:
```suggestion
            combinator.set_signals(None) # Clear signals
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+            try:
+                blueprint.add_circuit_connection("red", here, right)
+            except KeyError:
+                pass
+            try:
+                blueprint.add_circuit_connection("red", here, below)
+            except KeyError:
+                pass
+
+    print("Number of items added:", signal_count)
```
MyNameIsTrez: Maybe rename `signal_count` to `item_count`, since the value of `signal_count` is only used here and it is just being incremented after each new `item` is gotten from the `for item in items.raw:` loop? I don't think the fact you're calling `continue` in the loop for hidden items before you increment `signal_count` matters here, but maybe I'm wrong.

If you just want to print the number of items, though, printing `len(items.raw)` seems much more appropriate, since that means you can remove any mention of `signal_count` altogether?

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+            blueprint.entities.append(combinator)
+            # Reset the combinator
+            combinators_added += 1
+            y = combinators_added % COMBINATOR_HEIGHT
+            x = int(combinators_added / COMBINATOR_HEIGHT)
+            combinator.set_signals(None)  # Clear signals
+            combinator.tile_position = (x, y)
+            signal_index = 0
+
+    # Add the last combinator if partially full
```
MyNameIsTrez: Why is there no partially full check needed?
Is it guaranteed to be partially full?

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+![Results](./img/item_stack_size_output.png)
+
+This doesn't only applies to items either. All entities, instruments, signals, recipes, modules, and tiles are pulled from the emulated load process and then stored as *pickle* files in Draftsman for later use. Any script can be designed to be completely flexible across these categories; extra instruments in a new programmmable speaker, new module types only in certain machines, complete virtual signal lists for signal mappings, etc., all are interpreted correctly by Draftsman. By saving the data of the current mod configuration internally for later, this also means that you only have to update the data once, each time you change what mods you're using.
```
MyNameIsTrez: I'd either clarify on what a pickle file is, or omit the mention of it entirely, since most readers won't know what it is.

MyNameIsTrez:
```suggestion
This doesn't only applies to items either. All entities, instruments, signals, recipes, modules, and tiles are pulled from the emulated load process and then stored as *pickle* files in Draftsman for later use. Any script can be designed to be completely flexible across these categories; extra instruments in a new programmable speaker, new module types only in certain machines, complete virtual signal lists for signal mappings, etc., all are interpreted correctly by Draftsman. By saving the data of the current mod configuration internally for later, this also means that you only have to update the data once, each time you change what mods you're using.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Image to Blueprint Converter
+
+[This was something I made on a whim.](https://github.com/redruin1/factorio-draftsman/blob/main/examples/image_converter.py) It uses `Pillow` to load an image and converts it to a blueprint intended to be visible from the map view, all in less than 150 lines of code:
```
MyNameIsTrez:
```suggestion
[This was something I made on a whim.](https://github.com/redruin1/factorio-draftsman/blob/main/examples/image_converter.py) It uses the `Pillow` image library to load an image, and converts it to a blueprint intended to be visible from the map view, all in less than 150 lines of code:
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+![Portrait of Mike](./img/mike_stoklasa.png "A peculiar man with a comically large flask, colorized.")
+
+Many improvements could be made to this:
+
+* Dithering is only implemented on 1x1 tiles; I was having trouble adjusting the algorithm to multi-tile entities.
+* Multi-tile entities also fail to adjust the error metric property; palettization doesn't work as well when some pixels are of different size.
```
MyNameIsTrez:
```suggestion
* Multi-tile entities also fail to adjust the error metric property; palettization doesn't work as well when some pixels are of different sizes.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+![Portrait of Jay](./img/jay_bauman.png "Jay Bauman, notorious hack fraud, criminal at large. Armed and dangerous.")
+
+### Factorio Movie Player Resurrection
+
+A little while ago, while trawling for projects to use as examples for this very article, I came across the classic [Factorio Sandstorm](https://www.youtube.com/watch?v=mgfwwqwxdxY). A perfect project to adapt; or, it would have been, if I could actually load the map in-game! The original map version was still dated to the tender version `0.14.20`. In addition, a lot of the migrations that took place over such a long span of time broke the functionality of the script to convert the image frames into map data, as well as some functionality of the map itself, so just downloading an old version of Factorio and coercing it along wasn't going to be enough to revive the old save. For example, the `build.lua` script in charge of encoding still called the `automation-science-pack` and `logistic-science-pack` items `science-pack-1` and `science-pack-2`. That should give you some indication of how old the map was!
```
MyNameIsTrez:
```suggestion
A little while ago, while trawling for projects to use as examples for this very article, I came across the classic [Factorio Sandstorm](https://www.youtube.com/watch?v=mgfwwqwxdxY). A perfect project to adapt; or, it would have been, if I could actually load the map in-game! The original map version was still dated to the tender version of `0.14.20`. In addition, a lot of the migrations that took place over such a long span of time broke the functionality of the script to convert the image frames into map data, as well as some functionality of the map itself, so just downloading an old version of Factorio and coercing it along wasn't going to be enough to revive the old save. For example, the `build.lua` script in charge of encoding still called the `automation-science-pack` and `logistic-science-pack` items `science-pack-1` and `science-pack-2`. That should give you some indication of how old the map was!
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+It troubled me, seeing such an iconic piece of Factorio history fall into disrepair. So took the time to fix the issues and migrate it all the way to 1.1.57:
```
MyNameIsTrez:
```suggestion
It troubled me, seeing such an iconic piece of Factorio history fall into disrepair. So I took the time to fix the issues and migrated it all the way to 1.1.57:
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+| ![video of feel the beat](./img/cut_and_shrunk.gif) |
+| ------------------------ |
+| Pictured is part of [Feel the Beat](https://youtu.be/aLZQ-0dHbiU?t=38), keeping the tradition of Darude music videos. |
+
+Changing the build script could be done by hand, but the signal `raw-wood` that the machine internally used no longer exists in modern Factorio. In order to fix it, I replaced all occurrences of the signal with it in the map with `artillery-wagon` (since I knew it would be unique), and that map updating was done with a script running Draftsman code. I also added a number of other scripts for extracting the images from the source, as well as taking screenshots of the result and stitching them into an output video, which pretty easy in Python. I used said scripts to make the output above.
```
MyNameIsTrez:
```suggestion
Changing the build script could be done by hand, but the signal `raw-wood` that the machine internally used no longer exists in modern Factorio. In order to fix it, I replaced all occurrences of the signal with it in the map with `artillery-wagon` (since I knew it would be unique), and that map updating was done with a script running Draftsman code. I also added a number of other scripts for extracting the images from the source, as well as taking screenshots of the result and stitching them into an output video, which is pretty easy in Python. I used said scripts to make the output above.
```

submissions/factorio_draftsman/main.md
```diff
@@ -0,0 +1,211 @@
...
+
+I was also playing around with the idea of using Draftsman to make a blueprint-based version of the builder script instead of the console-based original method; this would allow you to see exactly what parts of the memory you are modifying by placing a grid aligned blueprint on top of them, and would also allow you to extend the memory of the machine by simply adding more blocks (the memory provided on the map is only enough for 4800 frames). I figured this was less important than updating what was there originally, especially when the code that *was* already there I knew worked. There could be other considerations as well; the amount of encoded data is pretty hefty, especially for blueprint strings which I suspect may be less dense than the console scripts, which are so large they are dumped in text files when built. Determining whether a blueprint method was actually feasible I figured could wait until I had some spare time further down the line.
```
MyNameIsTrez:
```suggestion
I was also playing around with the idea of using Draftsman to make a blueprint-based version of the builder script, instead of the console-based original method; this would allow you to see exactly what parts of the memory you are modifying by placing a grid aligned blueprint on top of them, and would also allow you to extend the memory of the machine by simply adding more blocks (the memory provided on the map is only enough for 4800 frames). I figured this was less important than updating what was there originally, especially when the code that *was* already there I knew worked. There could be other considerations as well; the amount of encoded data is pretty hefty, especially for blueprint strings which I suspect may be less dense than the console scripts, which are so large they are dumped in text files when built. Determining whether a blueprint method was actually feasible I figured could wait until I had some spare time further down the line.
```

Conor12345 (approved):

submissions/factorio_draftsman/main.md
redruin1: Draftsman that uses Python. This is just saying that modded entities are stored in the same place in Draftsman as vanilla ones; if you update your installation with a set of mods and then print `draftsman.data.entities.raw` it will display all the entities currently loaded, vanilla or modded. This mimics how it works in Factorio, where a modded entity is just added to the list of prototypes of that entity's type; a modded accumulator would just get added to the list of all accumulators, from which you can query via Draftsman script.

submissions/factorio_draftsman/main.md
redruin1: The `item_count` vs. `signal_count` was something I debated over changing; basically it depends on what you're looking at. If you're looking at it from the perspective of the items, then `item_count` makes more sense, but if you consider the fact that all of these eventually get resolved to constant combinator signals, `signal_count` makes more sense. 

The "hidden" check also complicates this somewhat. I'm a bit fuzzy on it at the moment, but I seem to remember it being possible that items could be hidden from the user, or belonged to entities that were hidden from the user; these items are useless to a script like this because the end user is only concerned with constructible entities. Thus, it makes sense to iterate over all the items, ignore the ones that are hidden, and then add everything else as signals to the constant combinators. This way of thinking also supports the use of `signal_count` over `item_count`, because the total number of signals in the resulting blueprint is not necessarily the total number of items in `items.raw`, which `item_count` could be conflated with.

In order to truly answer the question, I need to research if my assumption that items can be hidden is true, or if that was an assumption I made early on that I forgot to rigorously check; this was one of the first functioning scripts I wrote, so it's likely to have some holdovers from previous iterations. Let me get back to you on this.

submissions/factorio_draftsman/main.md
redruin1: Hmm, I suppose you're right actually. This is a bug, though a rare one; the final combinator will always be partially full unless the total number of item signals is perfectly divisible by 20 (or whatever the number of signal slots in the constant combinator is).
This could be fixed by checking the length of the underlying key array:
```python
# This is slightly obscure due to the way the underlying data is structured, but might still be legible enough
# You could also argue the opposite, that this is more illustrative of exactly how Draftsman emulates Factorio
if len(combinator.control_behavior["filters"]) > 0:
    combinator.id = "{}_{}".format(x, y)
    blueprint.entities.append(combinator)
```
Or, because the data is inputted front to back, we can check the first entry to see if it exists: if it does, then we know it has at least one entry that needs to be added:
```python
# This may also seem a little strange if one forgets that the data is added sequentially
if combinator.get_signal(0):
    combinator.id = "{}_{}".format(x, y)
    blueprint.entities.append(combinator)
```
Which one do you think is better?

redruin1: Actually, on second thought, I might extend the syntax of Draftsman to add a `signals` attribute to `ConstantCombinator`. It would alias `control_behavior["filters"]`, so it would use the same internal format, but it would be a much clearer shorthand, so you could write something like:
```python
if len(combinator.signals) > 0:
    # Add the combinator
```
That's probably the best way to go, though its not currently part of the syntax. Should I update Draftsman quickly to add this feature, or should we just use one of the two above?

submissions/factorio_draftsman/main.md
redruin1: Feel free to omit it; can be rewritten as `and then stored in Draftsman for later use.`

submissions/factorio_draftsman/main.md
MyNameIsTrez: To be honest, the only code I *really* understood was that last snippet, so if you could update Draftsman so that snippet is valid that'd be awesome :)

_drafts/main.md
```diff
@@ -0,0 +1,204 @@
...
+
+![Results](https://media.alt-f4.blog/ALTF4/61/item_stack_size_output.jpg)
+
+This doesn't only applies to items either. All entities, instruments, signals, recipes, modules, and tiles are pulled from the emulated load process and then stored as *pickle* files in Draftsman for later use. Any script can be designed to be completely flexible across these categories; extra instruments in a new programmmable speaker, new module types only in certain machines, complete virtual signal lists for signal mappings, etc., all are interpreted correctly by Draftsman. By saving the data of the current mod configuration internally for later, this also means that you only have to update the data once, each time you change what mods you're using.
```
MyNameIsTrez:
```suggestion
This doesn't only applies to items either. All entities, instruments, signals, recipes, modules, and tiles are pulled from the emulated load process and then stored in Draftsman for later use. Any script can be designed to be completely flexible across these categories; extra instruments in a new programmmable speaker, new module types only in certain machines, complete virtual signal lists for signal mappings, etc., all are interpreted correctly by Draftsman. By saving the data of the current mod configuration internally for later, this also means that you only have to update the data once, each time you change what mods you're using.
```

submissions/factorio_draftsman/main.md
redruin1: Will, do I'll push it as a patch to version `0.9.2` in a bit

submissions/factorio_draftsman/main.md
redruin1: Alright, it is necessary. Using the following code snippet on a vanilla install of Draftsman:
```python
for item_name in items.raw:
    item = items.raw[item_name]
    if "flags" in item and "hidden" in item["flags"]:
        print(item_name)
```
We get the following list:
```
loader
fast-loader
express-loader
rocket-part
coin
tank-machine-gun
vehicle-machine-gun
tank-flamethrower
artillery-wagon-cannon
spidertron-rocket-launcher-1
spidertron-rocket-launcher-2
spidertron-rocket-launcher-3
spidertron-rocket-launcher-4
tank-cannon
player-port
item-unknown
electric-energy-interface
linked-chest
heat-interface
linked-belt
infinity-chest
infinity-pipe
simple-entity-with-force
simple-entity-with-owner
burner-generator
```
A lot of these are just special entities that only exist in the editor, like infinity chests and loaders. Some are special components of larger entities, like the spidertron rocket launchers, and coin is a regular item that just isn't used. Thus, we do have to check each item to see if it's "valid" before proceeding.

I propose the following changes to make this more clear:
* Rename `signal_count` to `signals_added`, in the same vein as `combinators_added`
* Move `signals_added += 1` to after we set the signal to make it a little more procedural
* Change the final `print("Number of items added:", signals_added)` to `print("Total number of item signals added:", signals_added)` to make it more clear

submissions/factorio_draftsman/main.md
redruin1: Alright, Draftsman is patched, code side works. Keeping this unresolved until the article matches the changes.

submissions/factorio_draftsman/main.md
redruin1: Unresolving until the changes are made in the article.

_drafts/main.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Automatic Item Stack Sizes
+
+Oftentimes, for you want to figure out how much storage you need for a specific amount of an item. However, storage in Factorio is based on slots, not amounts, so the amount of storage is actually dependent on not only the amount you're trying to store, but also the stack size of that item. [You can design a circuit contraption](https://factorioprints.com/view/-LwH68dlvOEIKWkU5JqR) to determine the number of slots that you need for any input item, but you'd need a big cell of combinators documenting the item sizes. This is *boring* to make, and easily broken if a new item is added or the stack size is changed. Clearly, for such a simple and repetitive task, a script is better suited:
+
+```python
+# Create a N x 5 grid of connected constant combinators, with every item and their stack size
```
MyNameIsTrez:
```suggestion
# Create an N x 5 grid of connected constant combinators, with every item and their stack size
```

_drafts/main.md
```diff
@@ -0,0 +1,205 @@
...
+* **Draftsman is stable.** A rigorous suite of tests ensure that Draftsman behaves predictably and correctly (or at least correctly enough), with a code coverage of 100%. Draftsman is verified to work on the latest versions of Python 2 and 3, and is compatible with all versions of Factorio greater than 1.0, with the metrics to prove it.
+* **Draftsman is descriptive.** Draftsman enforces "Factorio-safety" as a core philosophy, which means that if a blueprint would result in an error on import in Factorio, an exception is thrown by Draftsman. Draftsman also tries to enforce "Factorio-correctness", which means that values that won't break, but are otherwise nonsensical, will raise warnings. Both errors and warnings are verbose, so that any problem with your script can be understood and resolved in seconds.
```
MyNameIsTrez:
```suggestion
* **Draftsman is descriptive.** Draftsman enforces "Factorio-safety" as a core philosophy, which means that if a blueprint would have an import error in Factorio, an exception is thrown by Draftsman. Draftsman also tries to enforce "Factorio-correctness", which means that values that won't break, but are otherwise nonsensical, will raise warnings. Both errors and warnings are verbose, so that any problem with your script can be understood and resolved in seconds.
```

_drafts/main.md
```diff
@@ -0,0 +1,205 @@
...
+* **Draftsman is close-to-source.** Draftsman bases all of its data off of [Wube's `factorio-data` repository](https://github.com/wube/factorio-data), which means that all entities are exactly as you would expect in-game, with no inconsistencies. This makes Draftsman up-to-date, makes updating Draftsman for future Factorio versions a breeze, and allows version control to monitor changes between Draftsman and Factorio, if any future breakage should occur.
+* **Draftsman supports mods.** Draftsman emulates [Factorio's data lifecycle](https://lua-api.factorio.com/latest/Data-Lifecycle.html) directly, meaning that the same loading process that happens when you launch the game is mimicked with a single Draftsman function. In addition to ensuring absolute accuracy to Factorio, this also makes it compatible with custom mod prototypes that can be accessed in script as if they were any other vanilla entity.
```
MyNameIsTrez:
```suggestion
* **Draftsman supports mods.** Draftsman emulates [Factorio's data lifecycle](https://lua-api.factorio.com/latest/Data-Lifecycle.html) directly, meaning that the same loading process that happens when you launch the game is mimicked with a single Draftsman function. In addition to ensuring absolute accuracy to Factorio, this also means that custom mod prototypes can be accessed from Draftsman as if they were any other vanilla entity.
```

_drafts/main.md
```diff
@@ -0,0 +1,205 @@
...
+
+Draftsman has custom classes designed around each individual entity and prototype, and is designed to work as seamlessly as possible with blueprint strings and other software. Draftsman converts to and from these prototypes when you import and export blueprint strings automatically, no extra steps required. You can import a blueprint string from Factorio as a `Blueprint` object, make any change you want, and then export that `Blueprint` object back into a string. Or, you can just make an entirely new `Blueprint` from scratch; Draftsman is designed to work around you, not for you to work around it.
+
+Draftsman also has support for custom `"EntityLike"` objects, most notably `Group` objects, that allow you to create custom constructs that can be inserted into blueprints for aid with clarity and compartmentalization. For example, you can make a design for a smelting block in a `Group`, and then you can place that block as many times as you want, whereever you want, rotated or flipped, essentially acting as a copy-paste tool within Draftsman.
```
MyNameIsTrez:
```suggestion
Draftsman also has support for custom `"EntityLike"` objects, most notably `Group` objects, that allow you to create custom constructs that can be inserted into blueprints for aid with clarity and compartmentalization. For example, you can make a design for a smelting block in a `Group`, and then you can place that block as many times as you want, wherever you want, rotated or flipped, essentially acting as a copy-paste tool within Draftsman.
```

_drafts/main.md
```diff
@@ -0,0 +1,205 @@
...
+            # Reset the combinator
+            combinators_added += 1
+            y = combinators_added % COMBINATOR_HEIGHT
+            x = int(combinators_added / COMBINATOR_HEIGHT)
+            combinator.set_signals(None) # Clear signals
+            combinator.tile_position = (x, y)
+            signal_index = 0
+
+    if len(combinator.signals) > 0:
+        # Add the last combinator if partially full
```
MyNameIsTrez:
```suggestion
    # Add the last combinator if partially full
    if len(combinator.signals) > 0:
```

